PR #300, opened 2014-06-05
Add logic for getting the current Timezone short name

jpitz:
When working on an application with a timezone drop down, there is no reliable way to ensure that you select the appropriate timezone.  You can use offset, but that does not guarantee a proper selection.  Using a combination of offset and the short name helps when you are working on a reduced list size.  The code below will return the short name for the current timezone.


concavelenz:
before anyone can look at this a CLA must be signed:

concavelenz:
No know CLA signature.

PR #304, opened 2014-07-03
Set XhrIoPool's superClass as final call

gingerik:
**What steps will reproduce the problem?**
Create a goog.net.XhrManager object with some default header(s) and a `minCount > 0`:
> var headers = new goog.structs.Map(); 
> headers.set('X-Custom-Header', 'This text should reach the server'); 
> var minCount = 1;
> var xhrManager = new goog.net.XhrManager(0, headers, minCount);

Send a request to the server
> xhrManager.send('req','server.php');

**What is the expected output? What do you see instead?**
The expected HTTP request message is (I removed any additional headers
that are not influenced by the sample code):
> GET /server.php HTTP/1.1
> X-Custom-Header: This text should reach the server

However, the transmitted message is:
> GET /server.php HTTP/1.1

**Cause**
This error is caused by the `goog.net.XhrIoPool` calling its superClass `goog.structs.PriorityPool` constructor, before storing the default headers. The constructor will immediately create `minCount` objects, lacking the default headers. (See also the comments on line 51 - 54 in `closure/goog/structs/prioritypool.js`.)


concavelenz:
This looks correct but it would be nice to have a regression test that shows the issue.

gingerik:
I agree. Is there some overview on the testing architecture? I'd enjoy writing a regression test if have a good starting point.

concavelenz:
Likely, I'm missing the obvious documentation.  However, any of the "test.js" files should be a good example (https://github.com/google/closure-library/blob/master/closure/goog/net/xhrio_test.js and its matching .html file for naive test runners).  The "goog.testing.jsunit" is the key dependency.  

There is also a chapter in the book.  
https://www.safaribooksonline.com/library/view/closure-the-definitive/9781449381882/ch15.html

gingerik:
OK, I managed to add two simple tests. Let me know what you think.

closure/goog/net/xhriopool_test.js
```diff
@@ -0,0 +1,50 @@
+// Copyright 2012 The Closure Library Authors. All Rights Reserved.
```
concavelenz: year

gingerik: updated `xhriopool_test.html` as well

concavelenz:
Looks good to me, can you update to fix the merge conflicts?

gingerik:
Done, @concavelenz. I had to update `lint_pull_request.sh` as well.

scripts/ci/lint_pull_request.sh
```diff
@@ -11,7 +11,7 @@ CHANGED_FILES=$(git diff --name-only --diff-filter=AM master..$CURRENT_BRANCH |
 
 if [[ -n "$CHANGED_FILES" ]]; then
   set -x
-  $USER_BASE/bin/gjslint --strict --jslint_error=all --exclude_files=deps.js $CHANGED_FILES
+  $USER_BASE/bin/gjslint --strict --jslint_error=all --exclude_files=deps.js,alltests.js $CHANGED_FILES
```
joeltine: You can revert this. Made this change in https://github.com/google/closure-library/commit/a785b1ffc7b71f78a7d2cbed8f7b48faa7934bb0.

gingerik: Done!

PR #319, opened 2014-07-20
Prefer vm instead of process binding evals

ruimarinho:
`require('vm')` should be preferred over `process.bindings('evals')`. The
latter has been removed in node 0.11.


nanaze:
@nicks  we cannot auto merge this at head. please reopen when ready.

PR #322, opened 2014-07-21
Perf improvements for sha1.js, sha2.js, and sha2_64b.js

coruus:
allocating additional memory (and incurring additional garbage
collections) unnecessarily. ~10% speed gain on Chrome 35.

goog/crypt/sha2.js: Port the block-size fast path from Sha1.update
to Sha2.update. Compute message schedule on the fly. Remove code
for testing for invalid inputs that is not present in Sha1.update.
~30% speed gain on Chrome 35.

goog/crypt/sha256_test.html: Remove testBadInput(); no similar test
is performed in sha1_test.html


dlg-yahoo:
There is a rebased and squashed version at https://github.com/google/closure-library/pull/391

meheffernan:
@dig-yahoo As I mentioned on Nov 13, it looks good to me.  Not sure if I'm supposed to explicitly approve it in the UI or not (I don't know github).  FWIW I just verified the performance results with the latest change on FF 34.0 and Chrome 40.0.  Looks good.

PR #337, opened 2014-08-28
Use changedTouches instead of targetTouches in style.getClientPosition

### LINKED ISSUE #41: goog.style.getClientPosition error with touchend event
This issue was imported from Closure Library's previous home at http://closure-library.googlecode.com

The link to the original issue is:
https://code.google.com/p/closure-library/issues/detail?id=588

fredj:
`targetTouches` is not defined for `touchend` or `touchcancel` events. `changedTouches` is always defined.

With the current implementation (after 116609a6de5a687a85299f3feeb6e044efa0d1a5) the position is always `[0, 0]` 

closes #41

### DIFF
--- a/closure/goog/style/style.js
+++ b/closure/goog/style/style.js
@@ -842,11 +842,10 @@ goog.style.getClientPosition = function(el) {
     var be = /** @type {!goog.events.BrowserEvent} */ (el);
     var targetEvent = el;
 
-    if (el.targetTouches && el.targetTouches.length) {
-      targetEvent = el.targetTouches[0];
-    } else if (isAbstractedEvent && be.getBrowserEvent().targetTouches &&
-        be.getBrowserEvent().targetTouches.length) {
-      targetEvent = be.getBrowserEvent().targetTouches[0];
+    if (el.changedTouches) {
+      targetEvent = el.changedTouches[0];
+    } else if (isAbstractedEvent && be.getBrowserEvent().changedTouches) {
+      targetEvent = be.getBrowserEvent().changedTouches[0];
     }
 
     return new goog.math.Coordinate(
--- a/closure/goog/style/style_test.js
+++ b/closure/goog/style/style_test.js
@@ -414,6 +414,10 @@ function testGetClientPositionEvent() {
 function testGetClientPositionTouchEvent() {
   var mockTouchEvent = {};
 
+  mockTouchEvent.changedTouches = [{}];
+  mockTouchEvent.changedTouches[0].clientX = 100;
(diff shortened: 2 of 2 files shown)


fredj:
Any change for this to get reviewed? Thanks

elemoine:
This change makes sense!

The `goog.style.getClientPosition` function may be used to return the client position of an event. In the case of a touch event it doesn't make sense to rely on the `targetTouches`.

This is the definition of `targetTouches` (from [mdn](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/targetTouches)):

« A TouchList listing all the Touch objects for touch points that are still in contact with the touch surface and whose touchstart event occurred inside the same target element as the current target element. »

So, on a touchstart, `targetTouches[0]` may not refer to the new touch point, but to a previous touch that is still in contact with the touch surface! So I think that currenly `goog.style.getClientPosition` may only work for the fist touchstart.

@fredj's patch fixes that problem. It also makes `goog.style.getClientPosition` work on `touchend` and `touchcancel` events, as mentioned in the issue description.

Potential caveat: when multiple touches are simultaneously added/removed to/from the touch surface then `goog.style.getClientPosition` doesn't make sense, because there are multiple "client positions" in that case. @fredj uses the touch object at index 0 in the `changedTouches` list. I think this is acceptable. @fredj what do you think about documenting this behavior in the comment block above the function?

fredj:
I'll add a comment, thanks

elemoine:
@fredj, in fact, with #405, `goog.style.getClientPosition` could look like this:

```js
goog.style.getClientPosition = function(el) {
  goog.asserts.assert(el);
  if (el.nodeType == goog.dom.NodeType.ELEMENT) {
    return goog.style.getClientPositionForElement_(
        /** @type {!Element} */ (el));
  } else {
    var targetEvent = el.changedTouches ? el.changedTouches[0] : el;
    return new goog.math.Coordinate(
        targetEvent.clientX,
        targetEvent.clientY);
  }
};
```

Indeed, with #405, if `el` is a `goog.events.BrowserEvent` then `el.clientX` and `el.clientY` should be correct already. So the only specific case is that of a native touch `Event`.

Thoughts?

elemoine:
See https://github.com/google/closure-library/compare/master...elemoine:41.

fredj:
@elemoine please create a pull request with your branch (and add `fixes #337` in the description)

elemoine:
See https://github.com/google/closure-library/pull/424.

joeltine:
https://github.com/google/closure-library/pull/424 is the canonical

PR #346, opened 2014-09-10
`goog.async.run` and `goog.async.nextTick` should be provided as member of `goog.async` namespace

KengoTODA:
This PR is same to #333, please check that PR for detail.

Note that I haven't updated `deps.js` in this branch. Please regenerate that file after you merge this to master.


concavelenz:
we won't be making this change.

PR #351, opened 2014-09-16
Update filedrophandler.js

supolik:
fix: Error accessing effectAllowed property in dragover event in IE

### DIFF
--- a/closure/goog/events/filedrophandler.js
+++ b/closure/goog/events/filedrophandler.js
@@ -196,7 +196,9 @@ goog.events.FileDropHandler.prototype.onElemDragOver_ = function(e) {
     e.stopPropagation();
     // Allow the drop on the drop zone.
     var dt = e.getBrowserEvent().dataTransfer;
-    dt.effectAllowed = 'all';
+    try {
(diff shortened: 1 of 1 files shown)


closure/goog/events/filedrophandler.js
```diff
@@ -196,7 +196,9 @@ goog.events.FileDropHandler.prototype.onElemDragOver_ = function(e) {
     e.stopPropagation();
     // Allow the drop on the drop zone.
     var dt = e.getBrowserEvent().dataTransfer;
-    dt.effectAllowed = 'all';
```
Ubehebe: Instead of the `try` ... `catch`, how about
```
if (dt.effectAllowed) {
  dt.effectAllowed = 'all';
}
```

Ubehebe:
Please also write a test in filehandler_test.js that fails at HEAD but passes under this patch.

concavelenz:
No CLA found: https://github.com/google/closure-library/blob/master/CONTRIBUTING

concavelenz:
No response.

PR #354, opened 2014-10-06
Made _GetInputByPath in closurebuilder.py use os.path.realpath.

camio:
A problem occured if 'closurebuilder.py' was run with relative path inputs
and the sources were absolute paths with symlinks within. 'os.path.abspath'
always resolves symlinks in the stem it adds to a relative path even though
it doesn't ever resolve symlinks for absolute paths.

The solution was to use 'os.path.realpath' instead which resolves all
embedded symlinks, providing an apples-to-apples comparison.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@
 # Name or Organization <email address>
 
 Google Inc.
+Stellar Science Ltd.
 Mohamed Mansour <hello@mohamedmansour.com>
 Bjorn Tipling <bjorn.tipling@gmail.com>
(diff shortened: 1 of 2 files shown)


nanaze:
Hi -- we need a CLA to be able to take this patch. We have no record for "camio"

See https://cla.developers.google.com

camio:
Hi -- There should be a CLA on file for all of Stellar Science. Is there any other information you need?

nanaze:
sorry, I do not see a CLA for Stellar Science.

nanaze:
Sorry, my bad. I have it.

PR #358, opened 2014-10-15
Use nodejs vm module instead of undocumented internal API call.

henvic:
It's also not available anymore on unstable (`0.11`).

For reference:
[utils: do not use internal Node API](https://github.com/caolan/nodeunit/commit/5ba1a94e92339ecdca322465abd741fb82974a40)


Ubehebe:
@nicks WDYT?

nicks:
lgtm! merge it! merge it!

PR #365, opened 2014-10-29
XhrIo: goog.base to the event listener constructor

mooyah:
Without this fix, I am unable to add event listeners to my goog.net.XhrIo object.


Ubehebe:
`goog.net.XhrIo.base` does the same thing as `goog.base`. In fact, 83c6a0b97bd16e23a8578ddb1a53d8e1a21fb4cf migrated most of Closure from `goog.base` to the `Foo.base` style.

PR #374, opened 2014-11-14
Fix goog.json.serialize treatment of functions

pauldraper:
`goog.json.serialize` sometimes returns invalid JSON.

```javascript
goog.json.serialize({foo:function(){}}) // returns '{}', good
goog.json.serialize([function(){},function(){}]) // returns '[,]', bad
```

While `'[,]'` is parseable by the lenient `goog.json.parse`, it is invalid JSON, and breaks for other parsers, e.g. `JSON.parse` or implementations in other languages.

After this change, the behavior is the same as the [ECMAScript standard](http://www.ecma-international.org/ecma-262/5.1/#sec-15.12.3) for `JSON.stringify` in these cases:

```javascript
goog.json.serialize({foo:function(){}}) // returns '{}'
goog.json.serialize([function(){},function(){}]) // returns '[null,null]'
```

---

There is additional nonstandard behavior, e.g. `goog.json.serialize(undefined)` is `''` rather than `undefined`. I would be happy to fix those as well, if we wanted it to adhere to that spec.

But at least after this change `goog.json.serialize` will output well-formed JSON.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,4 +14,5 @@ Ivan Kozik <ivan.kozik@gmail.com>
 Rich Dougherty <rich@rd.gen.nz>
 Chad Killingsworth <chadkillingsworth@missouristate.edu>
 Dan Pupius <dan.pupius@gmail.com>
+Paul Draper <paulddraper@gmail.com>
 
--- a/closure/goog/json/json.js
+++ b/closure/goog/json/json.js
@@ -216,6 +216,7 @@ goog.json.Serializer.prototype.serializeInternal = function(object, sb) {
     case 'boolean':
       sb.push(object);
       break;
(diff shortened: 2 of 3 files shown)


joeltine:
Thanks for the heads up on this Paul et al.

The goog.json.serialize(undefined) case has changed since this was first opened.  In 35c9311042b95796d7b12f58cd2bec6086052f7e serializing undefined now returns 'null'. 

The fix to serialize a function property seems reasonable. Because this PR has potential to break a bunch of internal stuff and the PR has gotten out of sync with the current implementation, I'm not going to merge it directly. I'll work on a fix internally, and push it out crediting you.

Moving to https://github.com/google/closure-library/issues/491

Issue #318, opened 2014-07-17
TwoThumbSlider: valueThumb would be covered by extentThumb


eyakcn:
When dragging the valueThumb towards extentThumb, the valueThumb should have a 'z-index' value bigger than extentThumb. And vice versa.

Give valueThumb a z-index value when start dragging it, then remove the z-index value when start dragging extentThumb.

joeltine:
PR stagnated.

PR #377, opened 2014-11-17
refs #318 fix zIndex of valueThumb and extentThumb

eyakcn:
Before:
image

After:
image


joeltine:
No follow up on PR request.

PR #394, opened 2015-01-08
Date symbols for rw_RW: Corrected from Gicuransi to Gicurasi (May) and changed other short names

nzeyimana:
Date symbols for rw_RW


mihnita:
Hello,

Thank you very much for your contribution.
Unfortunately, it cannot go in "as is", because the file is generated.

The way to fix this is more complicated, but also more rewarding :-)
Here is how:

The data comes from Unicode Common Locale Data Repository (CLDR, http://cldr.unicode.org/)
In this case, from this area of the CLDR Survey Tool: http://st.unicode.org/cldr-apps/v#/rw/Gregorian/

Short summary on how CLDR works:
 * accepts locale data from contributors (native speakers, linguists) using the Survey Tool
 * the contributors can submit, then argument pro / against certain changes
 * the contributors can submit supporting documents (dictionaries, country academy, newspapers, etc.)
 * the contributors vote to make the final decision
 * a new version of CLDR is released every 6 months

You can either
* become a CLDR contributor (more complicated, but also more power, recommended if you want to contribute longer time)
* you can just file a bug at http://unicode.org/cldr/trac (very lightweight, you don't even have to create an account, can even do it anonymously). Of course, some adding some official supporting documents to the bug can help, if is not an obvious mistake

CLDR is then used by closure, and also by the ICU libraries (http://site.icu-project.org)
It all trickles down into Mac OS X, Android, Linux, and hundreds of other projects.

Yes, it is more difficult than a github fix, and will take longer.
But you get the satisfaction that you made things better for all the Kinyarwanda speakers in the world.
Between http://cldr.unicode.org/#TOC-Who-uses-CLDR- and http://site.icu-project.org/#TOC-Who-Uses-ICU- you can see that pretty much everybody uses CLDR and / or ICU.

Thank you very much,
Mihai

PR #398, opened 2015-01-14
fixing type annotations so end-to-end compiles

soycode:
Hi - I work on uProxy, a Google Ideas project, and have been incorporating https://github.com/google/end-to-end to support encryption. I found today that I could no longer build some dependencies, and traced it back to these type annotations - specifically, without this change I get the message "ERROR - Bad type annotation. type not recognized due to syntax error" when trying to build end-to-end (which itself builds closure-library).

If you have some other process to get these changes in let me know, but for now I figured I'd make the pull request to at least notify the right people. Thanks!


MatrixFrog:
We've already fixed this internally, we just need to do a MOE push.

soycode:
K great, thanks!

PR #401, opened 2015-01-18
Correct documentation around goog.userAgent.product.ANDROID, CHROME.

sevaseva:
Add some tests with modern browsers' useragent strings.

### DIFF
--- a/closure/goog/useragent/product.js
+++ b/closure/goog/useragent/product.js
@@ -50,13 +50,14 @@ goog.define('goog.userAgent.product.ASSUME_IPAD', false);
 
 /**
  * @define {boolean} Whether we know at compile-time that the product is an
- *     Android phone.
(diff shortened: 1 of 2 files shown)


sevaseva:
Given where the browsers are, it may make sense to add goog.userAgent.product.CHROMIUM and deprecate goog.userAgent.product.CHROME = CHROMIUM. Developers would care whether it is based on Chromium 99.9% of the time, not whether it is Google Chrome.

In addition to Android WebViews, Opera on Linux, Windows, OSX and Android is based in Chromium.

If something like that will be accepted, I may be able to contribute that.

concavelenz:
no CLA found, which is necessary before anyone looks at this:

sevaseva:
Thanks for discovering this, but this should have been closed by now. I submitted this internally a while ago per request from the team.

PR #403, opened 2015-01-22
Prevent IE from using the dataset property in goog.dom.dataset.

### LINKED ISSUE #396: goog.dom.dataset.set does not work on IE11 sometimes
IE11 has support of 'dataset' property on Element.
But sometimes it is not working properly, while old setAttribute works fine.
goog.dom.dataset.set uses 'dataset' attribute first, so it is also working bad.

The case is:
1. Have some CSS class, which uses data-attr:
        .inputHint:before
        {
            content: attr(data-content);
        }
2. Change data-attr from js:
            var testSpan = document.getElementById("spanTest");
            test.testSpan['content'] = 'some val';
            // test.testSpan['content'] = '';
Nothing changes in IE11 here.

Test: http://jsbin.com/pumejubowu

nicksay:
In IE (up to and including IE 11), setting `element.dataset` in JS does not
propagate the values to CSS.  This breaks CSS expressions such as
`content: attr(data-content)` that would otherwise work.  Ensure that IE uses
the `element.setAttribute` code path instead.

Closes #396.


nicksay:
Updated to avoid using workarounds like `msPointerEnabled` in favor of a standard check, and removed the additional define to simplify the IE code path.

Ubehebe:
@nicksay You sent an internal CL for this, right? It'll appear on GitHub the next time someone does a push.

nicksay:
Yes, I'm keeping the external PR in sync for the benefit of the issue thread.  I'll close this once the internal CL is pushed.

nicksay:
This was committed as ff4954ce8dba70f9725909e537334646e23b6741

PR #410, opened 2015-02-15
Fix documentation in JSDoc comments

SandraShklyaeva:
The PR fixes some @code/@see tags in JSDoc that point to wrong methods. 

It also fixes usage of HTML in JSDocs. In the changed files, the html tags should be wrapped in <code></code>. Otherwise, it is treated as HTML formatting in comments.


concavelenz:
looks like there are conflict, do you want to update this?

SandraShklyaeva:
Yes, I'll resolve it shortly.

PR #417, opened 2015-02-24
in loadMany, return resulting deferred object to allow further manipulat...

bebbi:
...ion

### DIFF
--- a/closure/goog/net/jsloader.js
+++ b/closure/goog/net/jsloader.js
@@ -78,6 +78,16 @@ goog.net.jsloader.scriptsToLoad_ = [];
 
 
 /**
+ * The deferred result of loading the URIs in scriptsToLoad_.
+ * We need to return this to a caller that wants to load URIs while
+ * a deferred is already working on them.
+ * @type {!goog.async.Deferred}
+ * @private
+ */
+goog.net.jsloader.scriptLoadingDeferred_;
+
+
+/**
  * Loads and evaluates the JavaScript files at the specified URIs, guaranteeing
  * the order of script loads.
  *
@@ -94,6 +104,8 @@ goog.net.jsloader.scriptsToLoad_ = [];
  * @param {Array<string>} uris The URIs to load.
  * @param {goog.net.jsloader.Options=} opt_options Optional parameters. See
  *     goog.net.jsloader.options documentation for details.
+ * @return {!goog.async.Deferred} The deferred result, that may be used to add
+ *     callbacks
  */
 goog.net.jsloader.loadMany = function(uris, opt_options) {
   // Loading the scripts in serial introduces asynchronosity into the flow.
@@ -103,7 +115,7 @@ goog.net.jsloader.loadMany = function(uris, opt_options) {
   //
   // To work around this issue, all module loads share a queue.
   if (!uris.length) {
-    return;
+    return goog.async.Deferred.succeed(null);
   }
 
   var isAnotherModuleLoading = goog.net.jsloader.scriptsToLoad_.length;
@@ -111,8 +123,10 @@ goog.net.jsloader.loadMany = function(uris, opt_options) {
   if (isAnotherModuleLoading) {
     // jsloader is still loading some other scripts.
     // In order to prevent the race condition noted above, we just add
-    // these URIs to the end of the scripts' queue and return.
-    return;
+    // these URIs to the end of the scripts' queue and return the deferred
+    // result of the ongoing script load, so the caller knows when they
+    // finish loading.
+    return goog.net.jsloader.scriptLoadingDeferred_;
   }
 
   uris = goog.net.jsloader.scriptsToLoad_;
@@ -122,8 +136,10 @@ goog.net.jsloader.loadMany = function(uris, opt_options) {
     if (uris.length) {
       deferred.addBoth(popAndLoadNextScript);
     }
+    return deferred;
   };
-  popAndLoadNextScript();
+  goog.net.jsloader.scriptLoadingDeferred_ = popAndLoadNextScript();
+  return goog.net.jsloader.scriptLoadingDeferred_;
 };
 
 
--- a/closure/goog/net/jsloader_test.js
+++ b/closure/goog/net/jsloader_test.js
@@ -114,14 +114,18 @@ function testLoadMany() {
 
   // Load test #3 and then #1.
   window.test1 = null;
+  window.test4 = null;
   var testUrls1 = ['testdata/jsloader_test3.js', 'testdata/jsloader_test1.js'];
-  goog.net.jsloader.loadMany(testUrls1);
+  var result = goog.net.jsloader.loadMany(testUrls1);
 
   window.test3Callback = function(msg) {
     testCase.continueTesting();
 
     // Check that the 1st test was not loaded yet.
     assertEquals('verification object', null, window.test1);
+
+    // check that result has not fired yet
+    assertFalse('deferred has fired', result.hasFired());
 
     // Load test #4, which is supposed to wait for #1 to load.
     testCase.waitForAsync('testLoadMany');
@@ -134,7 +138,22 @@ function testLoadMany() {
 
     // Check that the 1st test was already loaded.
     assertEquals('verification object', 'Test #1 loaded', window.test1);
+
+    // all scripts loaded - verify that result has fired
+    assertTrue('deferred has fired', result.hasFired());
+
+    // on last script loaded, set variable
+    window.test4 = msg;
+
+    testCase.waitForAsync('testLoadMany');
   };
+
+  result.addCallback(function() {
+    testCase.continueTesting();
+
+    // verify that the last loaded script callback has executed
+    assertEquals('verification object', 'Test #4 loaded', window.test4);
+  });
 }
 
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@
 # Name or Organization <email address>
 
 Google Inc.
+gigmade ltd.
 Mohamed Mansour <hello@mohamedmansour.com>
 Bjorn Tipling <bjorn.tipling@gmail.com>
 SameGoal LLC <help@samegoal.com>


concavelenz:
Needs CLA: https://github.com/google/closure-library/blob/master/CONTRIBUTING

bebbi:
cla done

bebbi:
@concavelenz  why label cla:no? CLA is "signed and filed" March 10 (via email confirmation from google)

bebbi:
fixed author now matching signed CLA (re-write commit)

bebbi:
verified email in github too in case that's necessary

closure/goog/net/jsloader.js
```diff
@@ -119,8 +119,9 @@ goog.net.jsloader.loadMany = function(uris, opt_options) {
     if (uris.length) {
       deferred.addBoth(popAndLoadNextScript);
     }
+    return deferred;
   };
-  popAndLoadNextScript();
+  return popAndLoadNextScript();
```
joeltine: Add a @return type to loadMany.

bebbi: ok will do in a week or so.

joeltine:
Updates on this?

bebbi:
done. (rebased commit, annotated return and simplified code.) Synopsis for PR:

Fixes issue where caller doesn't know when load sequence is finished in loadMany
Fixes issue where it's impossible to parallelize script loading with multiple independent loadMany calls, e.g.: loadMany([A, B, C]) and loadMany([D, E, F])
Impact: since caller knows what's going on, there is no need for a goog.net.jsloader.scriptsToLoad_ queue. Clients can avoid the race condition mentioned.

closure/goog/net/jsloader.js
```diff
@@ -94,36 +94,26 @@ goog.net.jsloader.scriptsToLoad_ = [];
...
-  // sequence for client B, even though client A's scripts haven't all been
-  // loaded yet.
-  //
-  // To work around this issue, all module loads share a queue.
+  // The client receives a deferred to manage parallel loads.
   if (!uris.length) {
-    return;
-  }
-
-  var isAnotherModuleLoading = goog.net.jsloader.scriptsToLoad_.length;
```
joeltine: Ehhh, I think I'd prefer you not change the behavior of this method this radically. The original impl was designed with a specific use case in mind and breaking that assumption could cause breakages internally.

Can you revert this back to the original fix w/ only `@return` added?

bebbi: Ok. I had asked for feedback on the google group. It's a very specific use-case. The problem I see is it blocks other use-cases, as the caller has no control. Do you agree w/ some option to use/skip the queue? A param (opt_queue), or maybe an add'l func? Tx

joeltine: I think the addition is reasonable. I wouldn't make it the default though. Adding another field to `opt_options` is probably the best way to do it.  Take a look at the `goog.net.jsloader.Options` type.

That said, can you do that in a separate PR? This returning deferred change seems valuable enough on its own.

closure/goog/net/jsloader.js
```diff
@@ -94,36 +94,26 @@ goog.net.jsloader.scriptsToLoad_ = [];
...
-  goog.array.extend(goog.net.jsloader.scriptsToLoad_, uris);
-  if (isAnotherModuleLoading) {
-    // jsloader is still loading some other scripts.
-    // In order to prevent the race condition noted above, we just add
-    // these URIs to the end of the scripts' queue and return.
-    return;
```
bebbi: I’ll do as you suggest. However, a quick request for comment:
It seems to me that the code loses a tad elegance.
To support the original use-case, we now need to store active deferred to return it in the if (isAnotherModuleLoading) clause.
But since we always return deferreds now, that use-case could easily be handled by the caller.
I'm happy to code for it in loadMany, but as "with queue" is more specific than w/o (judged by the number of lines), it would seem more natural to use the option to enable it ("..plus one uri queue please") rather than disable it ("..oh, without that queue thing please").
I am aware that backwards compatibility can be a issue somewhere.
With that, could you confirm that you prefer it this way and I'll PR per your spec.

bebbi:
ok nevrmind, going to submit per spec. Just let me know in case you change decision

bebbi:
I'm seeing a linter issue for long line. Going to truncate all lines to 80 and re-submit

bebbi:
See #523 for the PR with skipQueue

closure/goog/net/jsloader.js
joeltine: >But since we always return deferreds now, that use-case could easily be handled by the caller.

I don't know if I agree with this. Yes, it's possible for the caller to manage it, but it would require passing around a deferred to anything that wants to chain loads. The way it is now, your only dependency is on `jsloader`, and invoking `loadMany` means you are always guaranteed to get the latest deferred regardless of where/how you call it. IMO, this seems nicer.

Re: backwards compatibility, this is the cost of maintaining a widely used library. You have to weigh the cost and potential payoff of a change. If we make this change and I break 200 tests, is it really worth me spending 3 days fixing them for this change? Especially when it's just as easy to just add an option to enable the functionality you want without breaking anyone.

closure/goog/net/jsloader.js
```diff
@@ -55,8 +55,8 @@ goog.net.jsloader.DEFAULT_TIMEOUT = 5000;
  * document: The HTML document under which to load the JavaScript. Default is
  *     the current document.
  * cleanupWhenDone: If true clean up the script tag after script completes to
- *     load. This is important if you just want to read data from the JavaScript
```
joeltine: I'm seeing this as 80 cols. Were you getting lint errors?

closure/goog/net/jsloader.js
```diff
@@ -122,15 +134,17 @@ goog.net.jsloader.loadMany = function(uris, opt_options) {
...
+  goog.net.jsloader.scriptLoadingDeferred_ = popAndLoadNextScript();
+  return goog.net.jsloader.scriptLoadingDeferred_;
 };
 
 
 /**
  * Loads and evaluates a JavaScript file.
  * When the script loads, a user callback is called.
- * It is the client's responsibility to verify that the script ran successfully.
+ * It's the client's responsibility to verify that the script ran successfully
```
joeltine: End w/ period.

closure/goog/net/jsloader.js
```diff
@@ -103,7 +113,7 @@ goog.net.jsloader.loadMany = function(uris, opt_options) {
   //
   // To work around this issue, all module loads share a queue.
   if (!uris.length) {
-    return;
+    return goog.async.Deferred.succeed(null);
```
joeltine: passing null isn't necessary.

joeltine: Actually, I'm guessing you're trying to emulate what `load` does? If so, that's fine.

closure/goog/net/jsloader.js
```diff
@@ -122,15 +134,17 @@ goog.net.jsloader.loadMany = function(uris, opt_options) {
     if (uris.length) {
       deferred.addBoth(popAndLoadNextScript);
     }
+    return deferred;
   };
-  popAndLoadNextScript();
+  goog.net.jsloader.scriptLoadingDeferred_ = popAndLoadNextScript();
+  return goog.net.jsloader.scriptLoadingDeferred_;
```
joeltine: Can you add a unit test that validates the deferred corresponds to the completion of loading all scripts?

bebbi:
done. Comment changes reset, my bad

closure/goog/net/jsloader_test.js
```diff
@@ -134,7 +138,24 @@ function testLoadMany() {
...
   };
+
+  result.addCallback(function() {
+    testCase.continueTesting();
+
+    // verify that the last loaded script callback has executed
+    assertEquals('verification object', 'Test #4 loaded', window.test4);
+  });
+
+
```
joeltine: kill newlines

bebbi: which ones? I've copied the writing style in that file, they're all over the place

joeltine: Between function definitions, yes. Not at the end of a function block though.

joeltine:
Good job on the test! I'm impressed. It's testing exactly what I expect and written in accordance with the surrounding style. Well done.

closure/goog/net/jsloader_test.js
bebbi: ah - just saw it now. tx

bebbi:
hey thanks. hope we've got it now (for today at least that's it for me..). tx for patience. I've rebased #523 on top of this one, will add test once agreed.

closure/goog/net/jsloader.js
```diff
@@ -78,6 +78,16 @@ goog.net.jsloader.scriptsToLoad_ = [];
 
 
 /**
+ * The deferred that is loading the uris in scriptsToLoad_
+ * We need to return this to a caller that wants to load uris while
+ * a deferred is already working on them
```
joeltine: end w/ period.

closure/goog/net/jsloader.js
```diff
@@ -111,8 +123,10 @@ goog.net.jsloader.loadMany = function(uris, opt_options) {
   if (isAnotherModuleLoading) {
     // jsloader is still loading some other scripts.
     // In order to prevent the race condition noted above, we just add
-    // these URIs to the end of the scripts' queue and return.
-    return;
+    // these URIs to the end of the scripts' queue and return the deferred
```
joeltine: Please use proper capitalization and end sentence w/ period.

bebbi: done. not sure what capitalization refers to. deferred? have changed sentence to work around.

joeltine: Oh, I'm sorry. I'm talking about English grammar. https://en.wikipedia.org/wiki/Capitalization

bebbi: I got that. But aside from the word 'deferred' (I tend to write instances in lower case but probably one can argue for upper case here), I didn't spot any case issues in other words that I added. Is it okay as amended?

joeltine: OHHH. Haha, I thought "these" started a new sentence. A side effect of using the "unified" view. Sorry for confusion.

joeltine:
Thanks!

bebbi:
Thanks to you. I hope we can add #523 too. I've checked all grammer and linting already :)

PR #418, opened 2015-02-25
Allow setting CLOSURE_BASE_PATH to the empty string

elemoine:
Currently, if `CLOSURE_BASE_PATH` is set to the empty string then it is considered as `undefined` and ignored by the `goog.findBasePath_` function. This PR fixes the problem by using the `goog.isDef` function for testing whether `CLOSURE_BASE_PATH` was set by the user.

### DIFF
--- a/closure/goog/base.js
+++ b/closure/goog/base.js
@@ -869,7 +869,7 @@ if (goog.DEPENDENCIES_ENABLED) {
    * @private
    */
   goog.findBasePath_ = function() {
-    if (goog.global.CLOSURE_BASE_PATH) {
+    if (goog.isDef(goog.global.CLOSURE_BASE_PATH)) {
       goog.basePath = goog.global.CLOSURE_BASE_PATH;
       return;
     } else if (!goog.inHtmlDocument_()) {


elemoine:
Let me explain why I want to `window.CLOSURE_BASE_PATH` to `''` (the empty string) in my application.

In my application I use Pyramid's [Cache Busting](http://docs.pylonsproject.org/projects/pyramid/en/master/narr/assets.html#cache-busting) for static resources, so URLs to static resources look like this: `http://www.example.com/static/c9658b3c0a314a1ca21e5988e662a09e/js/myapp.js`. Note the md5 hash in the URL.

For this to work I need to let the web framework (Pyramid) generate the static URLs (using `request.static_url`) when generating the `goog.addDependency` statements. So my `goog.addDependency` statements look like this:

```js
goog.addDependency('http://localhost:5000/proj/a1805ca3eeeb4733bd670f7492e9961e/js/appmodule.js', ['app'], ['ngeo']);
```

And for this to work I obviously need to set `window.CLOSURE_BASE_PATH` to the empty string.

I hope this makes sense, and I hope this explanation can help this patch be merged into master.

elemoine:
@MatrixFrog, this one should be quite easy for you :-)

PR #419, opened 2015-03-03
Use Object.keys instead of the pure js implementation in goog.object.getKeys

fredj:
Add a new goog.object.ASSUME_NATIVE_FUNCTIONS define to activate it.
The default is still to use the native js implementation.

Object.keys browser support: http://kangax.github.io/compat-table/es5/#Object.keys

### DIFF
--- a/closure/goog/object/object.js
+++ b/closure/goog/object/object.js
@@ -21,6 +21,14 @@ goog.provide('goog.object');
 
 
 /**
+ * @define {boolean} If true, JSCompiler will use the native implementation
+ * of object functions where appropriate and remove the unused
+ * pure JS implementation.
+ */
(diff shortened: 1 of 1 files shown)


elemoine:
@fredj have you see micro-benchmarks comparing the performance of both approaches? Just curious.

fredj:
No, I haven't

fredj:
See http://jsperf.com/object-keys-vs-hasownproperty for a speed comparison

joeltine:
Is there a justifiable reason for this change?

fredj:
Forgot about this pull request...
I don't need this anymore because I use `Object.keys()` directly (dropping support for IE < 8)

PR #420, opened 2015-03-05
Fix for early cancel in goog.fx.Dragger

mtsgrd:
An early cancel is triggered on the second drag start unless we set `this.dragging_ = false` after a first early cancel has occurred. Can someone please give their 2 cents on whether this is an appropriate fix or if I'm missing something?


joeltine:
Possible reasons EARLY_CANCEL is dispatched: disabled dragger, dragging with the right mouse button or releasing the button before reaching the hysteresis distance.

Please provide a repro case using https://jsfiddle.net/c2c6zvrk/ if you can demonstrate a bug.

PR #423, opened 2015-03-07
Fixes #422 by getting scrollTop from document.documentElement before document.body

### LINKED ISSUE #422: `goog.dom.getDocumentScroll` always returns 0 for a scrollTop (y coordinate) in IE on Windows Mobile
Here's how to simulate it:
 1. Create a huge container inside `document.body`, that huge so you can scroll the page right/bottom;
 2. Append something to that container, so it's not empty;
 3. Add `goog.events.EventType.SCROLL` event listener to a `window` object and debug `goog.dom.getDocumentScroll()` value inside a handler ([source code snippet](http://paste.ofcode.org/8UPsP3hiLsSnrSavYkDvLa));
 4. Open sample HTML page that loads JavaScript doing all the above in *IE11 browser on Windows Mobile* and start scrolling the page down/up (vertically), so scroll event handler is fired.

When debugging `goog.math.Coordinate` which is returned by `goog.dom.getDocumentScroll()`, you'll see that Y coordinate value is always 0.

However, when scrolling the page horizontally, X coordinate value is changing without an issue.<br /> This proves that the issue is only in identifying Y coordinate.

*Getting the issue on Windows Phone Prestigio and Nokia Lumia, both with Windows&reg; Phone 8.1 OS on IE11 browser.*

manakor:
In IE11 on Windows Phone `document.body.scrollTop` is always 0 when running in HTML5 Doctype. However, `document.documentElement.scrollTop` is updated correctly when scrolling the page vertically.

Fixes #422.

### DIFF
--- a/closure/goog/dom/dom.js
+++ b/closure/goog/dom/dom.js
@@ -607,7 +607,8 @@ goog.dom.getDocumentScroll_ = function(doc) {
     // The keyboard on IE10 touch devices shifts the page using the pageYOffset
     // without modifying scrollTop. For this case, we want the body scroll
     // offsets.
-    return new goog.math.Coordinate(el.scrollLeft, el.scrollTop);
+    return new goog.math.Coordinate(el.scrollLeft, goog.userAgent.MOBILE &&
+        doc.documentElement && doc.documentElement.scrollTop || el.scrollTop);
   }
   return new goog.math.Coordinate(win.pageXOffset || el.scrollLeft,
       win.pageYOffset || el.scrollTop);


concavelenz:
I was unable to find a CLA. There needs to be one before anyone can look at this.

manakor:
@concavelenz, there shall be one already. I am submitting a pull request to this repo not a first time. Would appreciate if you check once again (looking for my Github username or full name 'Nikita Sumeiko' in Google's CLA individual records).

concavelenz:
I was able to confirm the CLA. Thanks.

manakor:
Have anyone from Google team is planning to look at it? I have provided all the necessary information to simulate issue #422 on Windows Mobile.

Please let me know if I can help more somehow.

concavelenz:
We an internal team meeting tomorrow. I'll see if I can get some of these
issues unblocked.

joeltine:
This is currently hard to test because, sadly, no one has a Windows device around here.  Is this reproducible in desktop IE by chance... or maybe even an emulator?

One thing that's curious is that all the [info I can find on this issue seems to be for older doctypes only](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=IE+document.body.scrollTop). It seems odd this would fail *only* for a semi-modern version of IE, with HTML5 doctype, and on mobile.

joeltine:
I found some emulators hosted by Windows: https://msdn.microsoft.com/en-us/library/windows/apps/Mt162269.aspx. Is this the same version you have?

joeltine:
I spent some time on this today, and I just simply wasn't able to reproduce this on any of the emulation tools available to me. Emulating Windows phone w/ IE (regardless of version) always increments the scrollTop properly. I even fiddled with the doc modes, and all worked well. Perhaps this is something specific to the manufacturer? 

Beyond that, this fix would likely need to be made in the underlying `getDocumentScrollElement_`  method whose job it is to return the appropriate scroll element for every UA. That's the real root cause of this.

@manakor Please let me know if this is reproducible on something I can easily test without the specific device you mentioned. I'd be happy to take another look.

PR #425, opened 2015-03-09
Add package.json file to support NPM

ChadKillingsworth:
This commit adds a package.json file to the project to allow the Node Package Manager to utilize the project as a dependency. The package.json specifically sets the `private` option to `true` to prevent the project from being published to [npmjs.com](https://www.npmjs.com/). Rather, consumers should specify a git-commit URL for versioning. The `version` field is required by NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "google-closure-library",
   "description": "Google's common JavaScript library",
-  "version": "20150126.0.0",
+  "version": "20150315.0.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/google/closure-library.git"
--- a/package.json
+++ b/package.json
@@ -1,11 +1,17 @@
 {
   "name": "google-closure-library",
   "description": "Google's common JavaScript library",
-  "version": "1.0.0",
+  "version": "20150126.0.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/google/closure-library.git"
   },
+  "keywords": [
+    "javascript",
+    "library",
+    "goog",
+    "closure"
+  ],
   "author": "The Closure Library Authors",
   "license": "Apache-2.0",
   "bugs": {
--- a/package.json
+++ b/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "google-closure-library",
+  "description": "Google's common JavaScript library",
+  "version": "1.0.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/google/closure-library.git"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "google-closure-library",
+  "description": "Google's common JavaScript library",
+  "version": "1.0.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/google/closure-library.git"
+  },
+  "author": "The Closure Library Authors",
+  "license": "Apache-2.0",
```
nicks: you should probably set 'main' to do something reasonable? like maybe load the nodejs bootstrapper and return 'goog'?

manakor:
If Google Closure will be installed as node module under `node_modules/` dir, I am wondering how will you use it on the client? I bet your project static directory (where you store static assets) is not `node_modules/`, right? So you will not be able to reference to it like follows on your HTML page while working in development mode:
````html
<script src="/closure-library/closure/goog/base.js"></script>
````

Or you think making duplicate copies of the library stored in different directories makes sense? The one from `node_modules/` is required for a server usage, and the one in static directory for client usage?

When in development mode, I personally store `closure-library/` inside my static directory and access it for both client/server usage. And when in production, I compile and switch static directory to a production one what doesn't include `closure-library/` at all, but just a compiled files. However, for server usage it's still being required from my development static directory. And it doesn't force me to have duplicate copies.

package.json
ChadKillingsworth: I can - I just didn't think it would do anything useful and other library projects don't have one. https://github.com/angular/angular.js/blob/master/package.json

ChadKillingsworth:
>If Google Closure will be installed as node module under node_modules/ dir, I am wondering how will you use it on the client? I bet your project static directory (where you store static assets) is not node_modules/, right?

I've seen this handled a few different ways:
  1. A pre-build task copies the library over to the static resources folder.
  2. The local dev server has a route that maps `node_modules/closure-library` to some relative url to the project root.
  3. The project has a dev task that watches source files and compiles the project with `SIMPLE_OPTIMIZATIONS` for development automatically.

As for duplicate copies - that's an implementation issue and you would need to make the decision that makes most sense for you.

ChadKillingsworth:
@MatrixFrog This PR needs merged before closure-library can be published to NPM.

MatrixFrog:
Sending this internally with the MOE config update so that it'll get synced without someone having to do it manually on the next MOE push.

PR #429, opened 2015-03-15
Update the CLOSURE_IMPORT_SCRIPT method

AlastairTaft:
I have accepted the CLA.

The proposed change will allow the use of absolute file paths passed as the 'src' parameter instead of only paths relative to the base.js file. This is handy when trying to separate out my dependency files into multiple packages.

I can now do something like this in my deps.js file.
goog.addDependency(__dirname + '/pbx/db/Connection.js', ['pbx.db.Connection'], [], false);
goog.addDependency(__dirname + '/pbx/db/DataGroup.js', ['pbx.db.DataGroup'], ['pbx.db.Connection', 'pbx.db.DataGroup'], false);

### DIFF
--- a/closure/goog/bootstrap/nodejs.js
+++ b/closure/goog/bootstrap/nodejs.js
@@ -60,7 +60,7 @@ global.goog = {};
 global.CLOSURE_IMPORT_SCRIPT = function(src) {
   // Sources are always expressed relative to closure's base.js, but
   // require() is always relative to the current source.
-  require('./../' + src);
+  require(path.resolve(__dirname + '/../', src));
   return true;
 };
 


elemoine:
Can't you just set `CLOSURE_BASE_PATH` to the empty string? See https://github.com/google/closure-library/pull/418.

AlastairTaft:
I don't think the node CLOSURE_IMPORT_SCRIPT method uses anything to do with the base path? If it were to be modified to use the base path though I still don't think it would allow me to add dependencies from multiple modules though, unless they all matched up with the path relative to the base path (no absolute paths), which I think would force it to have some dependency on specific package file system structures, which is what I'm trying to solve.

joeltine:
I'm not too familiar with Closure in Node, but I'm a bit confused by this proposal. Shouldn't [depswriter.py](https://github.com/google/closure-library/blob/master/closure/bin/build/depswriter.py), when used correctly, just be auto-generating the correct relative paths for you? You shouldn't ever be manually editing your deps.js file.

Please re-open this issue with more context if you still think this is a valuable change.

Issue #51, opened 2014-04-15
goog.events.KeyHandler is not dispatching KEY event on arrow keys in Opera


nanaze:
This issue was imported from Closure Library's previous home at http://closure-library.googlecode.com

The link to the original issue is:
https://code.google.com/p/closure-library/issues/detail?id=568

PR #435, opened 2015-04-03
#51 - fix for Opera "keyCode" dispatching

SandraShklyaeva:
The PR fixes ticket #51 in general case. The ticket describes the problem for arrow keys only. Nevertheless, the problem appears for all keys.

P.S. Decided to resubmit a pull request for this completely to fix the mess I've done with my branches...


joeltine:
Fixed. See https://github.com/google/closure-library/issues/51

Issue #352, opened 2014-10-01
goog.array.reduce does not properly invoke native implmentation.  fix provided.


moagrius:
`goog.array.reduce` attempts to invoke the native method where possible, but fails to set the `previous` argument if `initialValue` is not provided.

reproduce:

```
// should be 3, but is NaN
goog.array.reduce([1,2], function(p, c){
  return p + c;
});
```

[the spec](http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.21) says "if initialValue is not provided..".  The MDN polyfill reads "not provided" by checking arguments.length.  Since that's not possible with goog.array.reduce accepting opt_obj as a 4th parameter, we can use goog.isDef.

So [line 354](https://github.com/google/closure-library/blob/master/closure/goog/array/array.js#L354) currently reads

```
return goog.array.ARRAY_PROTOTYPE_.reduce.call(arr, f, val);
```

instead use apply and an array of parameters determined by `val` being undefined or not.

```
return goog.array.ARRAY_PROTOTYPE_.reduce.apply(arr, goog.isDef(val) ? [f, val] : [f]);
```

Here's the completed replacement drop in (old line commented out):

```
var log = function(p, c){
  console.log(p, c);
  return p + c;
};

console.log('unmodified goog.array.reduce');
goog.array.reduce([1,2], log);

console.log('native array.prototype.reduce');
[1,2].reduce(log);

goog.array.reduce = goog.NATIVE_ARRAY_PROTOTYPES &&
  (goog.array.ASSUME_NATIVE_FUNCTIONS ||
    goog.array.ARRAY_PROTOTYPE_.reduce) ?
  function(arr, f, val, opt_obj) {
    goog.asserts.assert(arr.length != null);
    if (opt_obj) {
      f = goog.bind(f, opt_obj);
    }
    //return goog.array.ARRAY_PROTOTYPE_.reduce.call(arr, f, val);
    return goog.array.ARRAY_PROTOTYPE_.reduce.apply(arr,
      goog.isDef(val) ? [f, val] : [f]);
  } :
  function(arr, f, val, opt_obj) {
    var rval = val;
    goog.array.forEach(arr, function(val, index) {
      rval = f.call(opt_obj, rval, val, index, arr);
    });
    return rval;
  };

console.log('modified goog.array.reduce');
goog.array.reduce([1,2], log);

console.log('native array.prototype.reduce');
[1,2].reduce(log
```

nanaze:
opening issue

[...]

PR #437, opened 2015-04-03
patching goog.array.reduce to conform to spec

moagrius:
updating both prototypal and procedural versions per #352.  

unit test added.

maintains backward compatibility - all `goog.array` tests pass, tested in IE6, IE7, IE8, IE9, IE10, IE11, Chrome, FireFox, Safari, Opera.


joeltine:
See https://github.com/google/closure-library/issues/352

Issue #360, opened 2014-10-16
NPE in EnterHandler plugin when range is null


moagrius:
If the EnterHandler is registered, the enter key event is received, but no range is detected (range == null), an Error is thrown (Uncaught TypeError - cannot call 'isCollapsed' of null).

A simple null test would prevent the error from being thrown.

As is:

```
goog.editor.plugins.EnterHandler.prototype.deleteCursorSelectionW3C_ =
    function() {
  var range = this.getFieldObject().getRange();

  // Delete the current selection if it's is non-collapsed.
  // Although this is redundant in FF, it's necessary for Safari
  if (!range.isCollapsed()) {
  ...
```

Fix:

```
goog.editor.plugins.EnterHandler.prototype.deleteCursorSelectionW3C_ =
    function() {
  var range = this.getFieldObject().getRange();

  // Delete the current selection if it's is non-collapsed.
  // Although this is redundant in FF, it's necessary for Safari
  if (range && !range.isCollapsed()) {
  ...
```

The call to `goog.editor.range.getDeepEndPoint(range, true)` at the end of the method body may need to be moved to the conditional block as well (untested).

PR #438, opened 2015-04-03
adding truthy test to range in deleteCursorSelectionW3C_

moagrius:
Occasionally `deleteCursorSelectionW3C_` will throw an NPE when range is null, at `!range.isCollapsed()`


moagrius:
per #360

joeltine:
Thanks.

PR #440, opened 2015-04-05
goog.string.html errors in ECMASCRIPT5_STRICT

ChadKillingsworth:
When compiled with a language_in/out of ECMASCRIPT5_STRICT and with the ACCESS_CONTROLS warning guard, the caja libraries produce errors.

### DIFF
--- a/third_party/closure/goog/caja/string/html/htmlparser.js
+++ b/third_party/closure/goog/caja/string/html/htmlparser.js
@@ -68,7 +68,7 @@ goog.string.html.HtmlParser.Entities = {
   lt: '<',
   gt: '>',
   amp: '&',
-  nbsp: '\240',
+  nbsp: '\xA0',
   quot: '"',
   apos: '\''
 };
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
This is being deleted

MatrixFrog:
Sorry I was wrong. Not being deleted :(

third_party/closure/goog/caja/string/html/htmlparser.js
```diff
@@ -209,7 +209,6 @@ goog.string.html.HtmlParser.Elements = {
 /**
  * Regular expression that matches &s.
  * @type {RegExp}
- * @private
```
MatrixFrog: I think the linter will complain that anything ending in `_` has to be private. So we probably have to rename these as well.

MatrixFrog: Are these used anywhere besides this file and htmlsanitizer.js? If not we can probably make them `@package`.

ChadKillingsworth: No they are restricted to those two files. I'll update with `@package`.

ChadKillingsworth:
@MatrixFrog done

Issue #443, opened 2015-04-09
Can't listen for touchstart events on a draggable element


Kenny806:
In my app, I use goog.fx.Dragger to scroll an element in the page (the element is both the target and the handle). However, this element contains children that have click listeners, and on touch devices these listeners won't be fired because of the preventDefault() call in goog.fx.Draggeer.prototype.startDrag.

Despite having found a workaround, I think it would be nice to make it work by default. Having done some testing, I have found that listening for goog.events.EventType.TOUCHMOVE instead of goog.events.EventType.TOUCHSTART in the constructor seems to perform the same AND it enables the use of click listeners on the draggable element. 

I do not know about all the use-cases that goog.fx.Dragger was intended for (I do know that mine is probably not one of them), so maybe I am missing something and such a change is undesirable, but from the testing I have done everything seems to work just fine.

joeltine:
startDrag fires a goog.fx.Dragger.EventType.START event on touchstart. This change will affect that behavior by firing START on touchmove events instead of start. This is semantically incorrect and could break code relying on that assumption.

Beyond that, it looks like the code wasn't intended to fit your use case. It specifically calls preventDefault() to avoid bad user experiences like following a clicked link instead of dragging an element.

PR #444, opened 2015-04-09
goog.fx.Dragger listens for touchmove instead of touchstart

Kenny806:
This PR adresses issue #443 and makes goog.fx.Dragger listen to goog.events.EventType.TOUCHMOVE instead of goog.events.EventType.TOUCHSTART on the handle. This makes click listeners inside the handle element work on touch devices.


joeltine:
See https://github.com/google/closure-library/issues/443

Issue #415, opened 2015-02-20
goog.ui.DatePicker allows selection of dates in other months when ShowOtherMonths is false


illandril:
When ShowOtherMonths is off for goog.ui.DatePicker, the numbers for the dates in other months is hidden, but you can still select them by clicking on the blank spot where the date would show.

joeltine:
goog.ui is considered only in maintenance mode for serious issues. For a work around, you can just apply a similar fix that the PR was attempting to-do.  E.g.,

``` javascript
datepicker.setShowOtherMonths(false);
datepicker.setUserSelectableDataRange(
    goog.date.DateRange.thisMonth(datepicker.getDate()));
```

PR #445, opened 2015-04-09
A fix for the out-of-month day selection UI bug

fbsolo:
This proposed commit fixes the out-of-month day selection UI bug
identified by illandril:

#415


fbsolo:
4-12-15 / additional fixes:

Revised commit adds fixes for two Google editor bugs:

1. Originally, a <p> paragraph tag at the start of a highlighted Google Editor 'editMe' textarea would map to an extra blank line

 <br><br>

and this commit fixes that bug.

2. Originally, the Google Editor 'fieldContents' in editor.html would have an &nbsp when the page itself first loads. This commit fixes that bug.

PR #461, opened 2015-05-04
Fix issue #51 for the new Opera browser (Blink engine).

blueoctopus:
Issue #51: goog.events.KeyHandler is not dispatching KEY event on arrow
keys in Opera

This fix ensures that any Opera browser running under the Blink engine
(identified as goog.userAgent.WEBKIT in Closure Library) will be handled
the same way as Chrome is handled by goog.events.KeyHandler#handleEvent.


nanaze:
looks good, making slight edits

Issue #34, opened 2014-04-15
Bug: goog.date.Date and DST


nanaze:
This issue was imported from Closure Library's previous home at http://closure-library.googlecode.com

The link to the original issue is:
https://code.google.com/p/closure-library/issues/detail?id=599

blueoctopus:
I've tested this bug again and it seems to be fixed in the updated Google Chrome: Version 42.0.2311.135 (64-bit).

However it's still reproducible in Firefox (version 37.0.2) with time zone set to to "Brasilia Summer Time" (City: São Paulo - Brazil).

Quick copy/paste code to reproduces this issue in the console:

```javascript
goog.now = function() { return new Date(2013, 9, 20, 19).getTime(); };
new goog.date.Date().getDate(); // returns 19, should return 20!
```

PR #462, opened 2015-05-06
Fix DST bug in goog.date.Date.

blueoctopus:
Although this is fixed in the current Google Chrome, it's still
reproducible in the newst Firefox (version 37.0.2).

See issue #34 for more details.


joeltine:
Fixed see https://github.com/google/closure-library/issues/34

Issue #47, opened 2014-04-15
Bug in goog.fx.DragScrollSupport#constrainBounds_


nanaze:
This issue was imported from Closure Library's previous home at http://closure-library.googlecode.com

The link to the original issue is:
https://code.google.com/p/closure-library/issues/detail?id=577

PR #463, opened 2015-05-06
Fix bug in goog.fx.DragScrollSupport#constrainBounds_ (issue #47).

blueoctopus:


concavelenz:
Can you include a unit test that demostrates the problem?

joeltine:
I added a test.

PR #468, opened 2015-05-13
This PR fixes a bug that will "arbitrarily" delete text in the Google Editor. This usually happens when the user first starts using a fresh Google Editor.

fbsolo:
This PR fixes a bug that will "arbitrarily" delete text in the Google Editor. This usually happens when the user first starts using a fresh Google Editor.

"Soon" after the Google Editor

http://www.closurecheatsheet.com/demos/demos/editor/editor.html

first opens, it can sometimes delete text in either of the textareas if the user clicks in either of those textareas. This bug will disappear after "a little bit" of Google Editor use; this PR is a permanent fix.

### DIFF
--- a/closure/goog/demos/editor/editor.html
+++ b/closure/goog/demos/editor/editor.html
@@ -77,7 +77,28 @@ hooked up to a toolbar.</p>
 
   <script>
   function updateFieldContents() {
-    goog.dom.getElement('fieldContents').value = myField.getCleanContents();
+
+  // This function must call myField.getCleanContents() because
+  // getCleanContents() itself directly and indirectly calls
+  // other functions the editor needs. For whatever reason, the
+  // editor loads with
+  //
+  //	USING_LOREM 
+  //
+  // from
+  //
+  //	editor\command.js
+  //
+  // already set and this would overwrite the value in the lower
+  // 'fieldContents' textarea.
+
+    var getCleanContentsRetVal = myField.getCleanContents();
+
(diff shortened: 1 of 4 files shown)


joeltine:
You're going to need to provide concrete steps to reliably repro the problem before anyone can consider looking at your PR. Here's a template you can use: https://jsfiddle.net/c2c6zvrk/. Feel free to re-open the issue once you have a repro case.

concavelenz:
Add the very least this would need a unit test.

fbsolo:
John Lenz

Hello. Thank you for the recommendation re: Google Closure Library issue
#502 <https://github.com/google/closure-library/issues/502> and associated PR
#468 <https://github.com/google/closure-library/pull/468>.

To learn more about unit testing standards, I did some research, first
reading through the Closure Library website
<https://github.com/google/closure-library> and then extending past it. I
looked in

https://code.google.com/p/closure-library/issues/detail?id=110

https://en.wikipedia.org/wiki/Unit_testing


https://github.com/google/closure-library


https://developers.google.com/closure/library/

http://google.github.io/closure-library/api/

http://google.github.io/closure-library/source/closure/goog/demos/


but I never found a specific description. If you have an example URL that
explained the Google Closure Library unit testing standards, I will
proceed. If you'd like, I'm willing to write a unit testing readme document
for the library <https://github.com/google/closure-library> once I see the
expected unit test structure.

Thank you!

- Frank Solomon
________________

PR #475, opened 2015-06-01
Corrected date format for Portuguese (both locales)

rafasoares:
Short date formats in Portuguese (both in Portugal and Brazil) are expressed as dd/MM/yyyy (http://en.wikipedia.org/wiki/Date_format_by_country). I've updated the symbols to reflect that.


concavelenz:
This file is auto-generated based on data from http://cldr.unicode.org/

rafasoares:
Great. It seems everybody auto-genereates this data based on somewhere else... lol
Okay, I'll try and propose a change there.

But thanks, anyway.

Issue #485, opened 2015-06-26
goog.events.FileDropHandler throws error SCRIPT63335: "Unexpected call to method or property access" on IE10+


rikuayanokozy:
Some IE user reports a bug when dragging a file over the droppable area on IE10+, the dragover event returns an error accessing e.dataTransfer.effectAllowed: "Unexpected call to method or property access." (SCRIPT65535)

**What steps will reproduce the problem?**
1.open f12 debug toolbar
2.run goog/demos/filedrophandler.html on IE10+
3.drag a file over the droppable area
![How To Reproduce](https://cloud.githubusercontent.com/assets/10336620/8380813/7274d52c-1c5a-11e5-807c-4d023e711099.png)

**What version of the product are you using? On what operating system?**
Latest clone from github, windows 8 and IE11(**Our users which use IE10 also reported this problem**)

**Please provide any additional information below.**
It seems to be an IE bug (http://connect.microsoft.com/IE/feedback/details/811625/cant-get-datatransfer-effectallowed-set-in-another-window-raises-a-script65535-error).
![bug](https://cloud.githubusercontent.com/assets/10336620/8380814/72a364a0-1c5a-11e5-8bda-481fd9045cae.png)

I add a try-catch block to prevent this error as a temporarily solution.

PR #486, opened 2015-06-26
Fix the IE10+ SCRIPT65535 error in goog.events.FileDropHandler

rikuayanokozy:
See more at https://github.com/google/closure-library/issues/485

### DIFF
--- a/closure/goog/events/filedrophandler.js
+++ b/closure/goog/events/filedrophandler.js
@@ -196,7 +196,10 @@ goog.events.FileDropHandler.prototype.onElemDragOver_ = function(e) {
     e.stopPropagation();
     // Allow the drop on the drop zone.
     var dt = e.getBrowserEvent().dataTransfer;
-    dt.effectAllowed = 'all';
+    // Temporarily solution to prevent IE10+ SCRIPT65535 error, see more at https://github.com/google/closure-library/issues/485
+    try {
+        dt.effectAllowed = 'all';
+    } catch (e) {}
     dt.dropEffect = 'copy';
   }
 };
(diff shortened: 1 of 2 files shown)


rikuayanokozy:
I signed it!

concavelenz:
Can you add a test case that would fail without your change?

rikuayanokozy:
Um... I have not idea about construct a DataTransfer object in test.

joeltine:
You don't want to construct the actual DataTransfer object. You can just create a mock object. The existing test already does this. For example: https://github.com/google/closure-library/blob/eb26f425dbd99a70d2955d8fcc892f2fd0178acb/closure/goog/events/filedrophandler_test.js#L55

rikuayanokozy:
Thanks for ur help. I added test case.

rikuayanokozy:
Okay, I use defineProperty instead of defineSetter, now the test is compatible with IE8+. The test renamed to testEffectAllowedExceptionIsCaught, unneeded eventdispatches and assertions has been removed. And now indent with 2.

rikuayanokozy:
OK, I'll commit soon.
The event DRAGENTER is needed, because of [event dragenter set dndContainsFiles_](https://github.com/rikuayanokozy/closure-library/blob/f13fd37ca06a5df76b073e1dd7c3723825c24774/closure/goog/events/filedrophandler.js#L151-L155) and [onElemDrop_ need it](https://github.com/rikuayanokozy/closure-library/blob/f13fd37ca06a5df76b073e1dd7c3723825c24774/closure/goog/events/filedrophandler.js#L217).

joeltine:
The PR  LGTM minus the quotes thing. I'l just change it myself before merging.

PR #489, opened 2015-07-19
Array async utilities

alexg-dev:
Array async utilies.
New feature provides async implementation of:
goog.array.forEach
goog.array.map
goog.array.filter
goog.array.reduce
goog.array.find
goog.array.findIndex
goog.array.count
goog.array.some
goog.array.every

Also there are new functions:
goog.array.async.filterMap - Combines map() and filter() functionalities
goog.array.async.findAll - Search for all elements, that satisfy condition
goog.array.async.findIndexes - Search for all indexes

Code is fully commented and covered by test.


alexg-dev:
I signed it

concavelenz:
Can you explain your use case and why this is the best solution? Or the
benefit of your solution? It seems like there would be a more composable
solution

alexg-dev:
That's pretty simple. It's very useful in case, when you need to process big data array or to do 'hard' operations with it without user agent locking.
As another example, take a look on [this library](https://github.com/caolan/async).

concavelenz:
I think you have a misconception as to what goog.async.run does.  It doesn't throttle work in any way and  tries to run in as soon as possible (it was written to back goog.Promise).  Assuming you are trying to allow work "in the background" without blocking other operations, you need some form of task schedule that actually breaks up the work and has some concept of latency goals.  

If you would like continue with this, perhaps a design discussion on closure-library-discuss@ would be the best path forward.  Closing for now.

PR #490, opened 2015-07-21
Venezuelian currency bug fix

AlexanderZon:
Bs --> Bsf.


concavelenz:
Most of the i18n data is auto generated from CLDR (http://cldr.unicode.org/).  It looks like this file is not however but we should make sure it is consistent with the CLDR before merging.

joeltine:
CLDR still shows Bs from what I can see.

joeltine:
We generate these symbols directly from [CLDR](http://cldr.unicode.org/). Please file a change request directly against them at http://cldr.unicode.org/index/bug-reports. Once changed, it will eventually make its way into Closure.

PR #492, opened 2015-07-24
 Patching goog.array.reduce to confrom to ECMA spec; requires @param initialVal to be optional

moagrius:
see #352

### DIFF
--- a/closure/goog/array/array.js
+++ b/closure/goog/array/array.js
@@ -337,7 +337,7 @@ goog.array.map = goog.NATIVE_ARRAY_PROTOTYPES &&
  *     the value of the current array element, the current array index, and the
  *     array itself)
  *     function(previousValue, currentValue, index, array).
- * @param {?} val The initial value to pass into the function on the first call.
+ * @param {?=} val The initial value to pass into the function on the first call.
  * @param {S=} opt_obj  The object to be used as the value of 'this'
  *     within f.
  * @return {R} Result of evaluating f repeatedly across the values of the array.
@@ -346,20 +346,33 @@ goog.array.map = goog.NATIVE_ARRAY_PROTOTYPES &&
 goog.array.reduce = goog.NATIVE_ARRAY_PROTOTYPES &&
                     (goog.array.ASSUME_NATIVE_FUNCTIONS ||
                      goog.array.ARRAY_PROTOTYPE_.reduce) ?
-    function(arr, f, val, opt_obj) {
-      goog.asserts.assert(arr.length != null);
-      if (opt_obj) {
-        f = goog.bind(f, opt_obj);
+  function(arr, f, val, opt_obj) {
+    goog.asserts.assert(arr.length != null);
+    var params = [];
+    for(var i = 1, l = arguments.length; i < l; i++){
+      params.push(arguments[i]);
+    }
+    if (opt_obj) {
+      params[0] = goog.bind(f, opt_obj);
+    }
+    return goog.array.ARRAY_PROTOTYPE_.reduce.apply(arr, params);
+  } :
+  function(arr, f, val, opt_obj) {
+    goog.asserts.assert(arr.length > 0);
+    var value, length = arr.length, i = 0;
+    if (arguments.length > 2) {
+      value = arguments[2];
+    } else {
+      value = arr[0];
+      i = 1;
+    }
+    for (; i < length; i++) {
+      if(i in arr){
+        value = f.call(opt_obj, value, arr[i], i, arr);
       }
-      return goog.array.ARRAY_PROTOTYPE_.reduce.call(arr, f, val);
-    } :
-    function(arr, f, val, opt_obj) {
-      var rval = val;
-      goog.array.forEach(arr, function(val, index) {
-        rval = f.call(opt_obj, rval, val, index, arr);
-      });
-      return rval;
-    };
+    }
+    return value;
+  };
 
 
 /**
--- a/closure/goog/array/array_test.js
+++ b/closure/goog/array/array_test.js
@@ -281,6 +281,12 @@ function testArrayReduceOmitDeleted() {
   assertEquals(2, goog.array.reduce(a, scope.testFn, 0, scope));
 }
 
+function testArrayReducePreviousValuePatch(){
+  assertEquals(3, goog.array.reduce([1,2], function(p, c){
+    return p + c;
+  }));
+}
+
 function testArrayReduceRight() {
   var a = [0, 1, 2, 3, 4];
   assertEquals('43210', goog.array.reduceRight(a, function(rval, val, i, arr) {


closure/goog/array/array.js
```diff
@@ -346,20 +346,33 @@ goog.array.map = goog.NATIVE_ARRAY_PROTOTYPES &&
...
                     (goog.array.ASSUME_NATIVE_FUNCTIONS ||
                      goog.array.ARRAY_PROTOTYPE_.reduce) ?
-    function(arr, f, val, opt_obj) {
-      goog.asserts.assert(arr.length != null);
-      if (opt_obj) {
-        f = goog.bind(f, opt_obj);
+  function(arr, f, val, opt_obj) {
+    goog.asserts.assert(arr.length != null);
+    var params = [];
+    for(var i = 1, l = arguments.length; i < l; i++){
```
joeltine: I'm not a big fan of this loop. It enables a use of the API that's not intended (i.e., it allows an arbitrary number of arguments). 

I'd convert this logic to just reference "val" and "f" directly and pushing onto "params".

closure/goog/array/array.js
```diff
@@ -337,7 +337,7 @@ goog.array.map = goog.NATIVE_ARRAY_PROTOTYPES &&
  *     the value of the current array element, the current array index, and the
  *     array itself)
  *     function(previousValue, currentValue, index, array).
- * @param {?} val The initial value to pass into the function on the first call.
+ * @param {?=} val The initial value to pass into the function on the first call.
```
joeltine: Change the param name to "opt_val"

closure/goog/array/array.js
```diff
@@ -346,20 +346,33 @@ goog.array.map = goog.NATIVE_ARRAY_PROTOTYPES &&
...
+      params.push(arguments[i]);
+    }
+    if (opt_obj) {
+      params[0] = goog.bind(f, opt_obj);
+    }
+    return goog.array.ARRAY_PROTOTYPE_.reduce.apply(arr, params);
+  } :
+  function(arr, f, val, opt_obj) {
+    goog.asserts.assert(arr.length > 0);
+    var value, length = arr.length, i = 0;
```
joeltine: One "var" per line per variable please.

joeltine: BTW, when in doubt always refer to the google JS style guide: http://google.github.io/styleguide/javascriptguide.xml?showone=var#var

closure/goog/array/array.js
```diff
@@ -346,20 +346,33 @@ goog.array.map = goog.NATIVE_ARRAY_PROTOTYPES &&
...
+    if (arguments.length > 2) {
+      value = arguments[2];
```
joeltine: Don't use arguments. Just look for the existence of "val" directly.

closure/goog/array/array_test.js
```diff
@@ -281,6 +281,12 @@ function testArrayReduceOmitDeleted() {
   assertEquals(2, goog.array.reduce(a, scope.testFn, 0, scope));
 }
 
+function testArrayReducePreviousValuePatch(){
```
joeltine: Don't mention this is testing a patch in the test name.

closure/goog/array/array.js
```diff
@@ -346,20 +346,33 @@ goog.array.map = goog.NATIVE_ARRAY_PROTOTYPES &&
...
+    } else {
+      value = arr[0];
+      i = 1;
+    }
+    for (; i < length; i++) {
```
joeltine: I'm pretty sure this whole thing could be written much more simply.  From the [spec description](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce), it sounds like all providing initialValue does is [unshift](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift) the value onto the original array. I'd suggest trying that and using the same forEach logic that was there before.

closure/goog/array/array_test.js
```diff
@@ -281,6 +281,12 @@ function testArrayReduceOmitDeleted() {
...
+  assertEquals(3, goog.array.reduce([1,2], function(p, c){
+    return p + c;
+  }));
```
joeltine: Write a test that exercises the case where native reduce isn't available.

closure/goog/array/array_test.js
joeltine: The test name description should describe what this is testing. This is actually testing behavior when initialValue isn't provided.

closure/goog/array/array.js
moagrius: the problem here is that both "val" and "f" would be "undefined", which is a valid value - to properly meet spec, you need to rebuild using arguments, explicitly using length.

closure/goog/array/array.js
moagrius: Same issue here.  If val is undefined, it still qualifies.  If you try the changes you've suggested against any standard test of reduce, I think the specific problem will become obvious - LMK if you'd like more detail

moagrius:
@joeltine thanks very much for your comments.  I'll definitely make the changes you suggested where appropriate, and barring the 2 spots where I commented, I agree with everything you noted.

The undefined thing I mentioned in the line comments (in response to yours) gets a little tricksy and it's been a while since I wrote this initially (I think it's been something like 7-8 months since the original issue), so I'm probably not explaining it as well as I could since it's not fresh in my mind - if it's not immediately obvious, post back and I'll try to write up something more detailed and specific.

Thanks again

closure/goog/array/array.js
moagrius: Here's a couple pens that might help explain the issue more clearly:
http://codepen.io/moagrius/pen/RPqMVG?editors=001 (just inline comments since the lines here are getting crowded, no output)
http://codepen.io/moagrius/pen/jPQzrz?editors=001 (has a simple example showing the issue)

joeltine: "f" can never be undefined. It's of type "function" which is non-nullable. See [JavaScript Types](http://google.github.io/styleguide/javascriptguide.xml#JavaScript_Types).

If "val" is undefined, you just don't push it onto "params". E.g.,

``` javascript
if (goog.isDef(val)) {
  params.push(val);
}
```

closure/goog/array/array.js
joeltine: See above.

closure/goog/array/array.js
moagrius: i understand your point about f being not nullable, and that's fine - but the second point is still not correct - `undefined` is a valid value, it should be tested by array length...  that's to say that `fun(undefined, undefined)` is different than `fun()`, in terms of how the reduce spec is provided, as i understand it

joeltine: Got it now. It didn't click because using "undefined" as an initialValue seems pretty nonsensical to me, but I guess it's technically supported. 

Sorry for the confusion.

joeltine: Thinking about this more, supporting initialValue as optional doesn't seem worth the added complexity here. Doing all this arguments stuff is quite ugly and the old API is much more clear. I also noticed that "reduceRight" uses the same non-optional initialValue. The way it is currently is more consistent.

moagrius: Gotcha, I understand.  FWIW, the current implementation fails most basic use-cases of reduce, e.g.,

```
// sum an array of numbers
goog.array.reduce([1,2], function(a,b){
  return a+b;
});
```

^ Fails, whereas the native implementation (and my patch) returns 3, as expected.

Not the end of the world, but definitely not a drop in replacement for Array.prototype.reduce.

joeltine: In your scenario, you'd just pass 0 as initialValue.

moagrius: Cool

PR #303, opened 2014-07-02
Correct goog.structs.Pool minCount default

gingerik:
The JSDoc's state that `goog.structs.Pool`'s `minCount` is `1` by default.
However, the specific source file tells us that

> this.minCount_ = opt_minCount || 0;

PR #496, opened 2015-07-29
Fix docs on pool's default minCount

gingerik:
Regression in docs, see #303


joeltine:
Oops, looks like this commit was another victim of https://github.com/gingerik/closure-library/commit/889d16ae55c74aa0144d0be29b61243a9a49dcf7. Thanks for the update.

PR #330, opened 2014-08-08
math.Long#toString: Fix wrap around of intermediate int with radix 36

dcodeIO:
As of https://github.com/dcodeIO/Long.js/issues/7

```js
goog.math.Long.fromString("zzzzzz", 36).toString(36) == "-z141z5";
```

Which seems to be caused by an intermediate 32 bit integer wrap around in Long#toString. Fix and (basic) test attached.

PR #498, opened 2015-07-30
Use a smaller digit batch size when encoding in base 36

gfredericks:
There was a bug in the Integer#toString function for base 36, due to the
way it computes six digits at a time, and the fact that 36^6 > 2^31.
Since 35^6 < 2^31 this is only a problem for base 36, and can be fixed
by checking the radix and switching to five digits if it happens to be
36.

### DIFF
--- a/closure/goog/math/integer.js
+++ b/closure/goog/math/integer.js
@@ -261,7 +261,10 @@ goog.math.Integer.prototype.toString = function(opt_radix) {
   var result = '';
   while (true) {
     var remDiv = rem.divide(radixToPower);
-    var intval = rem.subtract(remDiv.multiply(radixToPower)).toInt();
+    // The right shifting fixes negative values in the case when
+    // intval >= 2^31; for more details see
+    // https://github.com/google/closure-library/pull/498
+    var intval = rem.subtract(remDiv.multiply(radixToPower)).toInt() >>> 0;
     var digits = intval.toString(radix);
 
     rem = remDiv;
--- a/closure/goog/math/integer_test.js
+++ b/closure/goog/math/integer_test.js
@@ -1691,3 +1691,10 @@ function testBigShift() {
   assertEquals('-591981510028266767381876356163880091648',
                a.negate().shiftLeft(97).toString());
 }
+
+// Regression test for
+// https://github.com/google/closure-library/pull/498
+function testBase36ToString() {
+  assertEquals('zzzzzz',
+               goog.math.Integer.fromString('zzzzzz', 36).toString(36));
+}
--- a/closure/goog/math/long_test.js
+++ b/closure/goog/math/long_test.js
@@ -1569,3 +1569,9 @@ function createTestToFromString(i) {
 for (var i = 0; i < TEST_BITS.length; i += 2) {
   goog.global['testToFromString' + i] = createTestToFromString(i);
 }
+
+// Regression test for
+// https://github.com/google/closure-library/pull/498
+function testBase36ToString() {
+  assertEquals('zzzzzz', goog.math.Long.fromString('zzzzzz', 36).toString(36));
+}
--- a/closure/goog/math/long.js
+++ b/closure/goog/math/long.js
@@ -380,7 +380,10 @@ goog.math.Long.prototype.toString = function(opt_radix) {
   var result = '';
   while (true) {
     var remDiv = rem.div(radixToPower);
-    var intval = rem.subtract(remDiv.multiply(radixToPower)).toInt();
+    // The right shifting fixes negative values in the case when
+    // intval >= 2^31; for more details see
+    // https://github.com/google/closure-library/pull/498
+    var intval = rem.subtract(remDiv.multiply(radixToPower)).toInt() >>> 0;
     var digits = intval.toString(radix);
 
     rem = remDiv;


gfredericks:
I signed it!

gfredericks:
Looks like this is a duplicate of #330

closure/goog/math/integer.js
```diff
@@ -253,9 +253,10 @@ goog.math.Integer.prototype.toString = function(opt_radix) {
     return '-' + this.negate().toString(radix);
   }
 
-  // Do several (6) digits each time through the loop, so as to
+  // Do several (5 or 6) digits each time through the loop, so as to
   // minimize the calls to the very expensive emulated div.
-  var radixToPower = goog.math.Integer.fromNumber(Math.pow(radix, 6));
+  var digitBatchSize = (radix == 36 ? 5 : 6);
```
joeltine: Can you please add a comment explaining why 5 is sometimes used? I.e., the issue with base 36.

closure/goog/math/integer_test.js
```diff
@@ -1649,3 +1649,9 @@ function testBigShift() {
   assertEquals('-591981510028266767381876356163880091648',
                a.negate().shiftLeft(97).toString());
 }
+
+function testBase36() {
```
joeltine: testBase36ToString

closure/goog/math/integer_test.js
```diff
@@ -1649,3 +1649,9 @@ function testBigShift() {
...
+  var x = goog.math.Integer.fromString('80000000',16);
```
joeltine: Can you provide a link to the github issue here?

closure/goog/math/integer_test.js
```diff
@@ -1649,3 +1649,9 @@ function testBigShift() {
...
+  assertEquals(x.toString(10),
+               goog.math.Integer.fromString(x.toString(36),36).toString(10));
```
joeltine: This test is hard to understand. Isn't there a simpler way to write this?

joeltine: E.g.,

```
assertEquals('zzzzzz', goog.math.Integer.fromString('zzzzzz', 36).toString(36));
```

gfredericks:
Before I put any more work into this, could you confirm you're interested in this approach over the one in #330? I've tested #330 pretty thoroughly and it seems to be correct (and much simpler than mine).

joeltine:
Oh, sorry. For some reason I was thinking that #330 was a fix for Long only. 

I do like the fix in #330 better. Can you change the PR to use `>>>` and can you also add the Long fix in this PR as well?

joeltine:
Actually, scratch that. Without a CLA or permission from the author of #330 I can't merge that change. We may just have to go with the 5 batch digits fix if we don't hear from him.

gfredericks:
Oh *wow* I didn't notice that #330 was about `Long` and not `Integer`. But in any case it's still true that the #330 fix applied to `Integer` fixes this bug.

Yeah the CLA issues were the main reason I wasn't sure how this should proceed.

Would you like an improved PR with the 5-digit fix applied to Long as well?

joeltine:
I'll try and contact the author. Hold off on doing anything just yet.

joeltine:
@gfredericks the author signed the CLA in https://github.com/google/closure-library/pull/330. Would you be willing to incorporate the Long fix into this PR as well?

gfredericks:
Will do, sounds good

gfredericks:
Pushed new commit that should address all the comments.

closure/goog/math/integer_test.js
```diff
@@ -1691,3 +1691,10 @@ function testBigShift() {
   assertEquals('-591981510028266767381876356163880091648',
                a.negate().shiftLeft(97).toString());
 }
+
+// Regression test for
+// https://github.com/google/closure-library/pull/498
+function testBase36toString() {
+  var x = goog.math.Integer.fromString('80000000',16);
```
joeltine: not used

closure/goog/math/integer_test.js
```diff
@@ -1691,3 +1691,10 @@ function testBigShift() {
...
+  assertEquals('zzzzzz', goog.math.Integer.fromString('zzzzzz', 36).toString(36));
```
joeltine: Is this 80 cols?

joeltine:
Where's the fix for Long?

gfredericks:
Oh I'm sorry I misinterpreted what you meant by "incorporate the Long fix". I'll add that too.

closure/goog/math/integer_test.js
gfredericks: no, I explicitly checked and it was somewhere around 50, but I must have checked in an extremely wrong fashion because it's actually 82.

gfredericks:
Pushed again

closure/goog/math/integer.js
```diff
@@ -261,7 +261,10 @@ goog.math.Integer.prototype.toString = function(opt_radix) {
   var result = '';
   while (true) {
     var remDiv = rem.divide(radixToPower);
-    var intval = rem.subtract(remDiv.multiply(radixToPower)).toInt();
+    // The right shifting fixes negative values in the case when
+    // intval >= 2^31; for more details see
+    // https://github.com/google/closure-library/pull/330
```
joeltine: Can you reference #498 instead, here and in Long. Thanks.

gfredericks: Sure, I thought 330 was more helpful, but will switch

closure/goog/math/long_test.js
```diff
@@ -1569,3 +1569,9 @@ function createTestToFromString(i) {
 for (var i = 0; i < TEST_BITS.length; i += 2) {
   goog.global['testToFromString' + i] = createTestToFromString(i);
 }
+
+// Regression test for
+// https://github.com/google/closure-library/pull/498
+function testBase36toString() {
```
joeltine: Capital "T" on "toString" here and in other test.

gfredericks: done

joeltine:
LGTM. \o/ this is a good fix. Thanks!

gfredericks:
☺

Issue #345, opened 2014-09-07
goog.structs.Map suggests that it works with the key "__proto__", but that is not the case


blueoctopus:
The @fileoverview comment of goog.structs.Map says:

> For simple keys such as numbers and strings, and where special names like \__proto__ are not a concern, consider using the lighter-weight utilities in goog.object.

Tested in Chrome 37:

```javascript
var map = new goog.structs.Map();
map.set('__proto__', 'ok');
map.get('__proto__'); // returns undefined, expected "ok"
```

PR #499, opened 2015-07-30
Remove mention of __proto__ being a safe key

joeltine:
Fixes https://github.com/google/closure-library/issues/345


joeltine:
I signed it!

PR #501, opened 2015-07-31
Box expand to include coordinate

ducklord:
Would be nice to be able to expand a box with just one coordinate. 
Changed the boundingBox() method to also use this function.


joeltine:
Can you address the CLA issue?

ducklord:
Not sure what went wrong, all commits has been made by me. But made from a few different machines. Had some issues in the merge to update to current master. When I got a mail about the CLA check failing I accepted that it was all me. So not sure why it still fails. Can you help me resolve the issue?

joeltine:
I'm guessing it's the commits using your real name instead of "ducklord" are those that are causing issues. Is there another email address you use when making those commits? If so, sign the CLA w/ it. Otherwise, I'd recommend just creating a fresh PR against the latest master with the ducklord commits.

ducklord:
Sorry for the trouble, a merge from master branch appeared to do the trick. I am unfortunately not fluent in git juggling yet. :)

closure/goog/math/box.js
```diff
@@ -174,6 +170,16 @@ goog.math.Box.prototype.expandToInclude = function(box) {
   this.bottom = Math.max(this.bottom, box.bottom);
 };
 
+/**
+ * Expand this box to include the coordinate.
+ * @param {goog.math.Coordinate} coord The coordinate to be included inside the box.
```
joeltine: This should be non-nullable type {!goog.math.Coordinate}.

joeltine: Please ensure 80 col limit (I can't tell if it is).

joeltine:
Can you give me an example use case where this method is more convenient than what's already in the API?

ducklord:
Currently it is not possible to expand an already existing box using a coordinate, only using another box. If I should expand a box using current API, I would have to create a new fake box, with the coordinate. The function that currently takes a list of coordinates that has been refactored, creates a new box. I want to reuse that bounding box, and expand it at a later time.

closure/goog/math/box.js
```diff
@@ -174,6 +170,17 @@ goog.math.Box.prototype.expandToInclude = function(box) {
   this.bottom = Math.max(this.bottom, box.bottom);
 };
 
+/**
+ * Expand this box to include the coordinate.
+ * @param {!goog.math.Coordinate} coord The coordinate to be included
+ * inside the box.
```
joeltine: indent 4. See examples throughout file.

In the future, you should try running the [closure linter](https://github.com/google/closure-linter) on your PR files. Will speed things up :). Eventually, this will be an automated step in Travis for all PRs, but for now you can run manually.

ducklord:
Sorry about that, did not notice that when multi line, there should be 4 space indentation. What setting should be added to the linter to test for this? Default settings does not report an error on the missing 4 spaces. Yes, I do use the linter. :) And yes would be awesome if the test ran automaticly, automagic is always awsome. :)

joeltine:
OOPS. My bad. I don't *think* the linter catches this particular jsdoc issue.

joeltine:
Thanks for the contribution!

PR #503, opened 2015-08-02
Add type information to goog.global referenced objects

ChadKillingsworth:
The `xhr` object created via goog.global was an unknown type and thus blocked disambiguation on properties named "open" or "send".

Fixes google/closure-compiler#1054

### DIFF
--- a/closure/goog/base.js
+++ b/closure/goog/base.js
@@ -41,6 +41,14 @@ var COMPILED = false;
  */
 var goog = goog || {};
 
+/**
+ * Forward declare type information. Used to assign types to goog.global
+ * referenced object that would otherwise result in unknown type references
+ * and thus block property disambiguation.
+ */
+goog.forwardDeclare('Document');
+goog.forwardDeclare('XMLHttpRequest');
+
 
 /**
  * Reference to the global context.  In most cases this will be 'window'.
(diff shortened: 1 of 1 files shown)


closure/goog/base.js
```diff
@@ -1396,6 +1396,7 @@ goog.loadFileSync_ = function(src) {
   if (goog.global.CLOSURE_LOAD_FILE_SYNC) {
     return goog.global.CLOSURE_LOAD_FILE_SYNC(src);
   } else {
+    /** @type {XMLHttpRequest} */
     var xhr = new goog.global['XMLHttpRequest']();
```
joeltine: I'm wondering if we can just kill the bracket notation here instead.

concavelenz:
You would also need to add a "goog.forwardDeclare" for this type, this is there to support projects using minimal externs.

closure/goog/base.js
ChadKillingsworth: The other option would be to increase the bracket usage:

    xhr['open']('get', src, false);

This is also occurring on [Line 850](https://github.com/google/closure-library/blob/master/closure/goog/base.js#L850) which doesn't use brackets.

ChadKillingsworth:
Added the needed `goog.forwardDeclare` calls and type casts to more objects.

PR #504, opened 2015-08-03
goog.json improvement

Leoperd87:
goog.json improvement
(https://developer.chrome.com/extensions/contentSecurityPolicy#relaxing-eval)

### DIFF
--- a/closure/goog/json/json.js
+++ b/closure/goog/json/json.js
@@ -25,12 +25,29 @@ goog.provide('goog.json.Serializer');
 
 
 /**
- * @define {boolean} If true, use the native JSON parsing API.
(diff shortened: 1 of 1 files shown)


Leoperd87:
I signed it!

joeltine:
Can you please be more explicit about what exactly this change intends to accomplish? A unit test would help as well.

Leoperd87:
My improvement mean that developer should select by his own how should work goog.json (same as goog.array). Because some environments didn't allow eval by security reasons.

concavelenz:
There are number of problems with this patch (it is a breaking change, it introduces a define in the wrong namespace, etc).  Can you be specific as to the problem you were facing and what you motivation was for this change.

PR #505, opened 2015-08-03
Wrong catch property name

vlkous:
Compiler output:
```
>> /var/folders/pl/w4db0j4d5gq8zk1lt6tgr34c0000gn/T/1438592089178.1658/bower_components/closure-library/closure/goog/events/filedrophandler.js:205: WARNING - Redeclared variable: e
>>     } catch (e) {
>>              ^
>> 
>> 0 error(s), 1 warning(s), 94.3% typed
```


joeltine:
Thanks. I actually am making this change when merging the original PR into the internal VCS. It should be out today or tomorrow.

vlkous:
Ok thx

PR #506, opened 2015-08-03
Allow setting CLOSURE_BASE_PATH to the empty string

fredj:
See #418 

This change got clobbered in 889d16a

### DIFF
--- a/closure/goog/base.js
+++ b/closure/goog/base.js
@@ -864,7 +864,7 @@ if (goog.DEPENDENCIES_ENABLED) {
    * @private
    */
   goog.findBasePath_ = function() {
-    if (goog.global.CLOSURE_BASE_PATH) {
+    if (goog.isDef(goog.global.CLOSURE_BASE_PATH)) {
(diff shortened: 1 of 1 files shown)


elemoine:
https://github.com/google/closure-library/commit/889d16ae55c74aa0144d0be29b61243a9a49dcf7 removes other changes to browservent.js we rely on!! I really don't understand how the development of closure-library works. External patches get merged and they are then reverted with such "force current state of Google VCS" commit!? How can we prevent this from happening?

elemoine:
In any case, I agree with my commit being merged into master again.

joeltine:
The TL;DR is things got really out of sync, and the easiest/quickest way to get into a good state was to force an external push. Up until recently Closure has only been a volunteer based effort. People didn't have a lot of time to spend untangling big messes. 

It's difficult keeping both an external and internal version of a large library like Closure in sync, especially considering they're not even the same VCS. Going forward though, this shouldn't happen again as Closure is formally staffed for some external support.

PR #515, opened 2015-08-07
Remove circular link to the Github repo

gterzian:


gterzian:
Signed it :memo:

joeltine:
Thanks!

concavelenz:
I'm not sure I would have merged this.  If you get the source bundle from another source, it is useful to know where the definitive source is.

gterzian:
@concavelenz That's a good point, and I wonder how many people would find themselves in such a position. The Google code page ranks higher in search results, and it only contains a link to the Github repo, so I think the possibility for confusion is very small. 

I found it quite annoying for this link to appear in the README, which in my opinion is all about briefly presenting the project in the best possible way(hence the PR). For those users would still might wonder what the "definitive source" is, perhaps some info could be added elsewhere, say, in the license file?

joeltine:
Ya, I didn't really consider that. I work with the GH source so much, it's easy to forget there can be others.

Reverted above.

Issue #500, opened 2015-07-31
goog.math.Integer: Division of large numbers causes an infinite loop


gfredericks:
As best I can tell, any number bigger than 10^309 is unprintable.

Printing a number requires serializing to base 10 which requires dividing by 1,000,000.

On dividing any such large number, the division algorithm creates an approximate answer using doubles.

The maximum representable double is somewheres around 10^308.

And so one way or another the division algorithm gets into an infinite loop.

gfredericks:
The commit I referenced (above) from my own fork seems to fix this problem (based on my extensive tests), but might not be the best performance-wise.

It ought to be strictly better than the current situation though, since it actually works and will be a fast check for not-very-large numbers.

Should I make a PR out of that, or wait for someone to do something with more of an eye on performance?

joeltine:
I'm not too familiar with this code. Trying to see if I can get someone more appropriate to take a look.

joeltine:
Unfortunately, I couldn't find any volunteers, so you're going to force me to brush up on my bit arithmetic now :).

It'd be very helpful if you could provide a breakdown explanation of what your algorithm does, and also provide some example test cases that demonstrate the issue.

gfredericks:
I will absolutely do that. I suppose I may as well upgrade this to a pull request while I'm putting in that effort.

gfredericks:
Okay I created #518, which has inline comments in the algorithm, and includes a regression test that can double as code to reproduce the problem. Let me know if there's anything I've left unaddressed.

PR #518, opened 2015-08-08
Degrade to naive division for very large numbers

gfredericks:
This is a fix for https://github.com/google/closure-library/issues/500
that could be rather more efficient, but it at least works rather than
causing an infinite loop.

Includes a regression test that causes an infinite loop without the fix.


closure/goog/math/integer_test.js
```diff
@@ -1640,6 +1640,13 @@ function testBigMultiply() {
   b = goog.math.Integer.fromString('-90689586573473848347384834');
   assertEquals('3573527027965969111849451155845960',
                a.multiply(b).toString());
+
+  // regression test for https://github.com/google/closure-library/issues/500
+  a = goog.math.Integer.fromString('1309694057237595253599107372802963115672745889511394745914382575458603399133344113551247898397218008347946268071052521756365987457287765181216375778104775434575839412393743854289058');
```
joeltine: 80 col limit plz: 

http://google.github.io/styleguide/javascriptguide.xml#Multiline_string_literals

closure/goog/math/integer.js
```diff
@@ -589,6 +643,13 @@ goog.math.Integer.prototype.divide = function(other) {
     return this.divide(other.negate()).negate();
   }
 
+  // Have to degrade to slowDivide for Very Large Numbers, because
+  // they're out of range for the floating-point approximation
```
joeltine: Link to github issue here or in the slowDivide jsdoc

closure/goog/math/integer.js
```diff
@@ -566,6 +566,60 @@ goog.math.Integer.carry16_ = function(bits, index) {
   }
 };
 
+/**
+ * Returns this Integer divided by the given one.
+ * NOTE: Both this and the given Integer MUST be positive.
```
joeltine: You can get rid of the "NOTE:" text.

closure/goog/math/integer.js
```diff
@@ -566,6 +566,60 @@ goog.math.Integer.carry16_ = function(bits, index) {
...
+ * @param {goog.math.Integer} other The Integer to divide this by.
```
joeltine: s/this/"this"/

joeltine: So goog.math.Integer is nullable by default, but the code doesn't seem to compensate. Line 593 would throw a runtime error if "other" was ever actually null.

closure/goog/math/integer.js
```diff
@@ -566,6 +566,60 @@ goog.math.Integer.carry16_ = function(bits, index) {
   }
 };
 
+/**
+ * Returns this Integer divided by the given one.
```
joeltine: This is the place to add a high level overview of what this method does, and why it exists.

closure/goog/math/integer.js
```diff
@@ -566,6 +566,60 @@ goog.math.Integer.carry16_ = function(bits, index) {
...
+ * NOTE: Both this and the given Integer MUST be positive.
+ * @param {goog.math.Integer} other The Integer to divide this by.
+ * @return {!goog.math.Integer} This value divided by the given one.
+ * @private
+ */
+goog.math.Integer.prototype.slowDivide_ = function(other) {
+  // This could be a lot better algorithmically, it's just a first an
```
joeltine: Can move comment like this up into the comments above the method.

closure/goog/math/integer.js
```diff
@@ -566,6 +566,60 @@ goog.math.Integer.carry16_ = function(bits, index) {
...
+    total2 = total.add(multiple);
+    if (total2.lessThanOrEqual(this)) {
+      res = res.add(twoPower);
+      total = total2;
+    }
+    multiple = multiple.shiftRight(1);
+    twoPower = twoPower.shiftRight(1);
+  }
+  return res;
+}
```
joeltine: semicolon.

closure/goog/math/integer.js
```diff
@@ -589,6 +643,13 @@ goog.math.Integer.prototype.divide = function(other) {
     return this.divide(other.negate()).negate();
   }
 
+  // Have to degrade to slowDivide for Very Large Numbers, because
+  // they're out of range for the floating-point approximation
+  // technique used below.
+  if(this.bits_.length > 30) {
```
joeltine: Missing space before "("

joeltine: On a side note, running your PRs through gjslint would probably be helpful: https://github.com/google/closure-linter

closure/goog/math/integer.js
gfredericks: > is nullable by default

does this mean all the code in the codebase is supposed to allow null arguments, or that I've accidentally indicated `other` can be null, or something else?

joeltine: Check out nullable and non-nullable types at http://google.github.io/styleguide/javascriptguide.xml#JavaScript_Types. For the most part, non-primitive types are nullable by default without explicitly specifying "!" in the type annotation.

Unfortunately, you can't just make this non-nullable as the calling method also (incorrectly) is passing a nullable Integer. You'll just have to code defensively for null.

gfredericks: But if I'm reading the code correctly, the `divide` method will crash on its first line (calling `.isZero`) if its argument is null, is that correct? So if `slowDivide` is only called by `divide` its argument should never be null?

And what does coding defensively mean here? What should this method do if it receives a null argument?

joeltine: Yes, you're right, `divide` would crash and it's unfortunate it's set up that way :(. However, each method should be considered in isolation and abide by its annotated type contract. You never know how or where this method will be used in the future. 

The right thing to do here is to make both the `divide` and `slowDivide`'s `other` params non-nullable. Considering this would completely break people's JS if they passed null right now anyway, it should be a safe change. Can you do that? Thanks.

gfredericks: Absolutely, that's exactly what I was thinking as well.

closure/goog/math/integer.js
gfredericks: Sorry about the style issues; I couldn't get the closure-linter to install. Hopefully there aren't more issues.

gfredericks:
Okay, I think all the comments are addressed.

closure/goog/math/integer_test.js
```diff
@@ -1640,6 +1640,26 @@ function testBigMultiply() {
...
+                                   '88954503636036324930799550973655515139183' +
+                                   '49748613352617850');
+  assertEquals('3656604243822473465231854689622824622925110708289864374262496' +
+               '6011714112304568802330367011373499705061131532809869274041668' +
+               '6403682258958556744594010497630468513209942389833344943567584' +
+               '9711887715782124066353351659885600959610314392971398216556707' +
+               '9177967386189614814272449310787401834779505072200615609456609' +
+               '84075750394399200851742167944560567661447926674910485300',
+               a.multiply(b).toString());
+
```
joeltine: kill newline

joeltine:
Thanks a lot for the PR. Hopefully this helps you and/or other people in the future. I'll merge it tomorrow.

OOC, is this division algorithm "standard"? The first result on Google is [one that involves subtraction](http://stackoverflow.com/questions/5284898/implement-division-with-bit-wise-operator). Why this one over others?

gfredericks:
I made up the algorithm myself. I think the difference between mine and the one you linked to (which I wasn't aware of) is that I didn't have to figure out the details of how the signed representation of `Integer` works (in particular how it works across the different 32-bit chunks). E.g., how do I determine the `bitCount` for a given Integer? I don't know.

PR #520, opened 2015-08-10
Fix textarea for Microsoft Edge (unit tests now pass)

anwerso:


anwerso:
I signed it!

concavelenz:
This is a good change, the only question I have is if we should invert the logic here, so we are only checking for old IE:

!(goog.userAgent.IE && !goog.userAgent.isDocumentModeOrHigher(11))

anwerso:
I thought about inverting logic too. In that case textarea will also work in Presto (no need to extra condition).

concavelenz:
@anwerso   Were you going to invert the condition? The lint failure should have been fixed separately.

PR #523, opened 2015-08-12
Add jsloaderskipqueue

bebbi:
Per discussion in #417, this commit enables a caller to skip the queue.
This supports parallel loading of independent sequences of scripts

### DIFF
--- a/closure/goog/net/jsloader_test.js
+++ b/closure/goog/net/jsloader_test.js
@@ -157,6 +157,61 @@ function testLoadMany() {
 }
 
 
+// Test the loadMany function with skipQueue option for independent URL loading.
+function testLoadManySkipQueue() {
+  testCase.waitForAsync('testLoadManySkipQueue');
+
+  // The regular result loads test #3.
+  window.test3 = null;
+  var testUrls1 = ['testdata/jsloader_test3.js'];
+  var result = goog.net.jsloader.loadMany(testUrls1);
+
+  // The result with skipQueue option loads test #4, then test #1.
+  window.test1 = null;
+  window.test4 = null;
+  var optionSkipQueue = {
+    skipQueue: true
+  };
+  var testUrlsSkipQueue = ['testdata/jsloader_test4.js', 'testdata/jsloader_test1.js'];
+  var resultSkipQueue = goog.net.jsloader.loadMany(testUrlsSkipQueue, optionSkipQueue);
+
+  // Verify that result and resultSkipQueue are separate deferreds.
+  assertNotEquals('deferreds are not identical', resultSkipQueue, result);
+
+  window.test3Callback = function(msg) {
+
+    // Set variable when the regular result has finished loading.
+    window.test3 = msg;
+  };
+
+  window.test4Callback = function(msg) {
+    testCase.continueTesting();
+
+    // Check that the last URL in resultSkipQueue sequence was not loaded yet.
+    assertEquals('verification object', null, window.test1);
+
+    // Set variable when first listed URL in resulSkipQueue was loaded.
+    window.test4 = msg;
+
+    testCase.waitForAsync('testLoadManySkipQueue');
+  };
+
+  // Both results finish independently. Wait for both and then continue testing.
+  result.awaitDeferred(resultSkipQueue);
+
+  result.addCallback(function() {
+    testCase.continueTesting();
+
+    // Check that resultSkipQueue has loaded both variables.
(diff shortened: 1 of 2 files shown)


bebbi:
(I'll add a test when the code is agreed)

concavelenz:
Can you describe your use case and why you feel this is generally useful?

bebbi:
This addresses an issue where it's impossible to parallelize the loading of independent sequence of scripts with multiple loadMany calls - e.g.: loadMany([A, B, C]) and loadMany([D, E, F]) can not be parallelized even if D/E/F have no dependency on A/B/C.

The proposed code allows skipping the queue forced by loadMany and have caller-side management of dependencies.

bebbi:
Additionally, in terms of general usefulness, this commit allows a caller to manage their own loadMany queue however it suits them - i.e. this will make loadMany more generic and increase the number of possible use-cases it can support. Any feedback appreciated.

joeltine:
Can you add a test for this please? Thanks!

bebbi:
- added tests
- I couldn't reproduce the deps.js formatting using depswriter.py (License text etc), so I manually added the new files. Best practice feedback welcome.
- I noticed some capitalization issues in jsloader_test.js from my previous PR #417 . Do you prefer this commit to fix it or an addional one?

closure/goog/net/jsloader_test.js
```diff
@@ -118,6 +118,22 @@ function testLoadMany() {
...
+    skipQueue: true
+  };
+  var skipQueueResult = goog.net.jsloader.loadMany(skipQueueTestUrls, options);
+
+  // Verify that the skipQueueResult is a separate deferred.
+  assertNotEquals('deferreds are not identical', skipQueueResult, result);
+
+  // Check that the independent test was not loaded yet.
+  assertEquals('verification object', null, window.test5);
+
```
joeltine: Can you move these tests into their own `testLoadManySkipQueue` method?

joeltine:
>I noticed some capitalization issues in jsloader_test.js from my previous PR #417 . Do you prefer this commit to fix it or an addional one?

Style/formatting changes should always be separate. This reduces required mental bandwidth of the reviewer, and protects those changes from being rolled back in the more complicated change.

closure/goog/net/jsloader_test.js
bebbi: Yes that should work. Unless you object, I'd still keep 1 loadMany + 1 loadMany(skipQueue) as the testing scenario though, it seems to make most sense.

concavelenz:
I'm not yet convinced this is an appropriate addition.  Please don't merge this yet.

bebbi:
@concavelenz I just run into another use-case now.
loading many scripts required:
loadMany(manyScriptURLs),
and in another caller:
loading 2 small scripts for more time critical stuff:
loadMany(smallstuff, {skipQueue: true})
Now there's a chance I can finish number 2 early without too much cross-talk, as long as the skipQueue feature exists. Otherwise my main view is going to wait long for all other stuff to load. Does that make sense?

joeltine:
@concavelenz what are your concerns about this change? Being able to load discrete sets of dependencies in parallel seems like it could be useful. It's off by default, so there isn't much harm in enabling these use cases.

bebbi:
Updated: separate test, and an additional commit for fixing comment syntax in merged PR #417

bebbi:
any news on this?

concavelenz:
The real question isn't whether the functionality is useful but whether it introduces unneeded complexity, confusion and risk to this class and if it would be better in a private fork of the class.   I'd feel better if others were advocating for the useful this of this change.

bebbi:
I see. That's possibly an information dissemination problem then, I've tried discussing loadMany on closure google group before and have found helpful people but no user. github issues are probably a better place. Generally, would be good to establish a place to agree on features with owners before coding.

joeltine:
>I've tried discussing loadMany on closure google group before and have found helpful people but no user.

I see you started [a thread on closure-library-discuss](https://groups.google.com/forum/#!searchin/closure-library-discuss/jsloader/closure-library-discuss/H-UI0bwVEsY/az3vxMF1L0gJ) back in January, which was the right thing to do. Closure has only been maintained by volunteers up until this June. These types of design proposals should get more attention going forward.

joeltine:
Sorry @bebbi, just can't get consensus on this. I'd recommend you just use your own forked impl or overwrite loadMany in your code base.

PR #524, opened 2015-08-13
Added missing drag event to goog.fx.AbstractDragDrop.

Znegl:
Until now the drag event was only fired just before the drop event — not continuously while dragging.
Now it dispatches a drag event for every drag event it receives from goog.fx.Dragger, making the event dispatching behavior of the two classes identical.

### DIFF
--- a/closure/goog/fx/abstractdragdrop.js
+++ b/closure/goog/fx/abstractdragdrop.js
@@ -588,7 +588,7 @@ goog.fx.AbstractDragDrop.prototype.moveDrag_ = function(event) {
       event.clientY,
       x,
       y
-  ));
+      ));
 
   // Check if we're still inside the bounds of the active target, if not fire
   // a dragout event and proceed to find a new target.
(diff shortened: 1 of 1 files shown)


Znegl:
I signed it!

closure/goog/fx/abstractdragdrop.js
```diff
@@ -575,10 +575,23 @@ goog.fx.AbstractDragDrop.prototype.moveDrag_ = function(event) {
   var x = position.x;
   var y = position.y;
 
-  // Check if we're still inside the bounds of the active target, if not fire
-  // a dragout event and proceed to find a new target.
   var activeTarget = this.activeTarget_;
 
+  // Dispatch a drag event
+  this.dispatchEvent(new goog.fx.DragDropEvent(
```
joeltine: can you format this similar to the event on line 609?

closure/goog/fx/abstractdragdrop.js
```diff
@@ -575,10 +575,23 @@ goog.fx.AbstractDragDrop.prototype.moveDrag_ = function(event) {
   var x = position.x;
   var y = position.y;
 
-  // Check if we're still inside the bounds of the active target, if not fire
-  // a dragout event and proceed to find a new target.
   var activeTarget = this.activeTarget_;
 
+  // Dispatch a drag event
```
joeltine: Comment seems unnecessary

PR #526, opened 2015-08-14
Fixed IE Edge bug in goog.fx.Dragger

Znegl:
Changed the check for support for setCapture() in goog.fx.Dragger from user agent sniffing to feature detection to make it work in IE Edge, which doesn't support setCapture.


Znegl:
Oops. Wrong branch.

PR #527, opened 2015-08-14
Fixed IE Edge bug in goog.fx.Dragger

Znegl:
Changed the check for support for setCapture() in goog.fx.Dragger from user agent sniffing  to feature detection to make it work in IE Edge, which doesn't support setCapture.

### DIFF
--- a/closure/goog/fx/dragger.js
+++ b/closure/goog/fx/dragger.js
@@ -212,11 +212,11 @@ goog.tagUnsealableClass(goog.fx.Dragger);
  * @type {boolean}
  * @private
  */
-goog.fx.Dragger.HAS_SET_CAPTURE_ =
+goog.fx.Dragger.HAS_SET_CAPTURE_ = goog.isDef(
     // IE and Gecko after 1.9.3 has setCapture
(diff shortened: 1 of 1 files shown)


closure/goog/fx/dragger.js
```diff
@@ -212,11 +212,8 @@ goog.tagUnsealableClass(goog.fx.Dragger);
  * @type {boolean}
  * @private
  */
-goog.fx.Dragger.HAS_SET_CAPTURE_ =
-    // IE and Gecko after 1.9.3 has setCapture
-    // WebKit does not yet: https://bugs.webkit.org/show_bug.cgi?id=27330
-    goog.userAgent.IE ||
-    goog.userAgent.GECKO && goog.userAgent.isVersionOrHigher('1.9.3');
+goog.fx.Dragger.HAS_SET_CAPTURE_ = goog.isDef(
```
concavelenz: Can you leave a comment that blick/webkit don't have setCapture.

concavelenz:
Please leave a comment as to which browsers are known not to have setCapture and then squash your commits?

PR #529, opened 2015-08-16
Update PasteHandler for MS Edge

omarkilani:
The latest change from IE -> EDGE broke this.

### DIFF
--- a/closure/goog/events/pastehandler.js
+++ b/closure/goog/events/pastehandler.js
@@ -94,6 +94,7 @@ goog.events.PasteHandler = function(element) {
 
   if (goog.userAgent.WEBKIT ||
       goog.userAgent.IE ||
+      goog.userAgent.EDGE ||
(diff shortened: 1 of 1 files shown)


omarkilani:
I signed it!

(As Remember The Milk Inc.)

closure/goog/events/pastehandler.js
```diff
@@ -94,6 +94,7 @@ goog.events.PasteHandler = function(element) {
 
   if (goog.userAgent.WEBKIT ||
```
concavelenz: I wonder if we can just drop the alternate branch.

omarkilani: I guess it depends what Google's minimum supported versions are?

I'd happily drop it... :)

concavelenz: ok, lets drop it.

concavelenz:
going with this as is until we publish a browser support policy

PR #530, opened 2015-08-17
Fix InputHandler for MS Edge.

omarkilani:
The latest IE -> EDGE change broke InputHandler...

### DIFF
--- a/closure/goog/events/inputhandler.js
+++ b/closure/goog/events/inputhandler.js
@@ -81,7 +81,7 @@ goog.events.InputHandler = function(element) {
   //   event.
   // IE9 supports input events when characters are inserted, but not deleted.
   // WebKit before version 531 did not support input events for textareas.
-  var emulateInputEvents = goog.userAgent.IE ||
+  var emulateInputEvents = goog.userAgent.IE || goog.userAgent.EDGE ||
       (goog.userAgent.WEBKIT && !goog.userAgent.isVersionOrHigher('531') &&
           element.tagName == goog.dom.TagName.TEXTAREA);
 


omarkilani:
I signed it!

(As Remember The Milk Inc.)

joeltine:
I'm not seeing "omarkilani" as having signed the CLA. Can you sign it with this user name?

closure/goog/events/inputhandler.js
```diff
@@ -81,7 +81,7 @@ goog.events.InputHandler = function(element) {
   //   event.
   // IE9 supports input events when characters are inserted, but not deleted.
   // WebKit before version 531 did not support input events for textareas.
-  var emulateInputEvents = goog.userAgent.IE ||
+  var emulateInputEvents = goog.userAgent.IE || goog.userAgent.EDGE ||
```
concavelenz: Can you change this so that this is on by default and off for old IE and old Safari.

omarkilani:
@joeltine: I honestly don't know what that means -- we have a corporate CLA that incorporates a Google Apps group with a bunch of email addresses in it. My commit email is in that group, but it wasn't tied to my GitHub account -- I've added that email to GitHub too... Does that fix it?

@concavelenz Yup.

As an aside, I feel like people are going to spend a lot of time re-adding all the IE checks back for Edge -- I obviously don't know how much testing is being done internally over there but it may have been better  (?) to just s/IE/EDGE_OR_IE/ when EDGE was introduced and then removed the hacks as it was proven they were no longer required.

As far as I can see, Edge still has all of the baggage associated with IE (execCommand, etc) and most of the same bugs. The layout engine even seems to need the exact same CSS hacks as IE 11.

closure/goog/events/inputhandler.js
omarkilani: I'm not sure that's right given the comment in the header:

```
 *     IE doesn't have native support for input event. WebKit before version 531
 *     doesn't have support for textareas. For those browsers an emulation mode
 *     based on key, clipboard and drop events is used. Thus this event won't
 *     trigger in emulation mode if text was modified by context menu commands
 *     such as 'Undo' and 'Delete'.
```

Seems to say that emulation is only on for those specific browsers (IE [and Edge], WebKit before 531)... so the code seems clear in that case?

I guess that can be EDGE_OR_IE to save some characters?

concavelenz: No, your right, I was sleepy.

concavelenz:
You were the not the first to state that IE Edge seems closer to IE 11 than Webkit/Blink and we had a long debate about this internally to Google.  There were two camps those that were for following the MS recommendation of treating MS Edge as "webkit" and those that wanted to threat it as "ie".  What we now  is Edge as separate layout engine.  I would like to know what they actually changed to be closer to webkit.

omarkilani:
@concavelenz Well... they could have fooled me. I guess it's early days but unless there's some X-UA-Compatible setting that I'm missing... I don't see how it's anything like Blink/WebKit.

Not to mention the glacial JS engine on real world apps, but that's another story entirely. I may be slightly jaded... :smile:

PR #531, opened 2015-08-17
Fixes finish weekday dateformat.

jbnicolai:
For reference, see https://github.com/angular/angular.js/issues/11148.


jbnicolai:
It seems the travis-ci build is failing due to linting errors not introduced in this PR.

concavelenz:
The files are autogenerated from other sources, as noted at the start of the file.  As such we don't accept changes to these files as any such changes would be overwritten the next time the files are generated.  You would need to get the root source updated.

joeltine:
See similar comment in https://github.com/google/closure-library/pull/532

PR #532, opened 2015-08-17
Update local currency symbol for DKK

c960657:
The local currency symbol for DKK should be changed from ``kr`` to ``kr.`` (note the trailing period).

``kr.`` has always been the official Danish abbreviation, but the ``kr``variant has also existed in glibc until recently (perhaps this how it made its way into Closure initially):
https://sourceware.org/bugzilla/show_bug.cgi?id=17692

### DIFF
--- a/closure/goog/i18n/currency.js
+++ b/closure/goog/i18n/currency.js
@@ -284,7 +284,7 @@ goog.i18n.currency.CurrencyInfo = {
   'COP': [0, '$', 'COL$'],
   'CRC': [0, '\u20a1', 'CR\u20a1'],
   'CZK': [50, 'K\u010d', 'K\u010d'],
-  'DKK': [18, 'kr', 'kr'],
+  'DKK': [50, 'kr.', 'kr.'],
   'DOP': [2, '$', 'RD$'],
   'EGP': [2, '£', 'LE'],
   'ETB': [2, 'Birr', 'Birr'],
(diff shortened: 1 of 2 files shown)


joeltine:
We generate these symbols directly from [CLDR](http://cldr.unicode.org/). If it's incorrect there, please file a change request directly against them at http://cldr.unicode.org/index/bug-reports. Once changed, it will eventually make its way into Closure.

c960657:
[The change](http://unicode.org/cldr/trac/changeset/10669/trunk/common/main/da.xml) was implemented in CLDR about a year ago and included in [release 26](http://cldr.unicode.org/index/downloads/cldr-26) that was released in September 2014.

Shouldn't it have appeared in Closure by now, or do I just need more patience? :-)

c960657:
@joeltine [The change](http://unicode.org/cldr/trac/changeset/10669/trunk/common/main/da.xml) was implemented in CLDR about a year ago and included in [release 26](http://cldr.unicode.org/index/downloads/cldr-26) that was released in September 2014.

Shouldn't it have appeared in Closure by now, or do I just need more patience? :-)

joeltine:
It's possible it hasn't be regenerated in a while. I can check.

joeltine:
OK. It turns out these (currency.js and currencycodemap.js) are the few remaining files we don't auto-generate, and apparently whoever last updated, missed this one. 

Our i18n maintainer is aware of this discrepancy and will update it as soon as possible in the internal closure. It will then make it out externally shortly after. 

On a side note, we are currently working on a big refactoring of all the locale data generators so this shouldn't happen again!

joeltine:
Oops thought this was an issue, not a PR. Please track https://github.com/google/closure-library/issues/534 for the fix on this.

PR #539, opened 2015-08-25
Fixes #126: goog.dom.xml is broken on MSIE9+

### LINKED ISSUE #126: goog.dom.xml is broken on MSIE9
This issue was imported from Closure Library's previous home at http://closure-library.googlecode.com

The link to the original issue is:
https://code.google.com/p/closure-library/issues/detail?id=456

vincenta:
This commit fixes the issue #126: goog.dom.xml is broken on MSIE9.

It has been reported this issue also affects MSIE 10 and 11.

The issue is available here: https://github.com/google/closure-library/issues/126
Original bug report: https://code.google.com/p/closure-library/issues/detail?id=456

### DIFF
--- a/closure/goog/dom/xml.js
+++ b/closure/goog/dom/xml.js
@@ -22,6 +22,7 @@ goog.provide('goog.dom.xml');
 
 goog.require('goog.dom');
 goog.require('goog.dom.NodeType');
+goog.require('goog.userAgent');
 
 
 /**
@@ -39,6 +40,39 @@ goog.dom.xml.MAX_ELEMENT_DEPTH = 256; // Same default as MSXML6.
 
 
 /**
+ * Check for ActiveXObject support by the browser.
+ * @return {boolean} true if browser has ActiveXObject support.
+ * @private
+ */
+goog.dom.xml.hasActiveXObjectSupport_ = function() {
+  if (!goog.userAgent.IE) {
+    // Avoid raising useless exception in case code is not compiled
+    // and browser is not MSIE.
+    return false;
+  }
+  try {
+    // Due to lot of changes in IE 9, 10 & 11 behaviour and ActiveX being
+    // totally disableable using MSIE's security level, trying to create the
+    // ActiveXOjbect is a lot more reliable than testing for the existance of
+    // window.ActiveXObject
+    new ActiveXObject('MSXML2.DOMDocument');
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
+
+
+/**
+ * True if browser has ActiveXObject support.
+ * Possible override if this test become wrong in coming IE versions.
+ * @type {boolean}
+ */
+goog.dom.xml.ACTIVEX_SUPPORT = goog.userAgent.IE &&
+                               goog.dom.xml.hasActiveXObjectSupport_();
+
+
+/**
  * Creates an XML document appropriate for the current JS runtime
  * @param {string=} opt_rootTagName The root tag name.
  * @param {string=} opt_namespaceUri Namespace URI of the document element.
@@ -48,11 +82,7 @@ goog.dom.xml.createDocument = function(opt_rootTagName, opt_namespaceUri) {
   if (opt_namespaceUri && !opt_rootTagName) {
     throw Error("Can't create document with namespace and no root tag");
   }
-  if (document.implementation && document.implementation.createDocument) {
-    return document.implementation.createDocument(opt_namespaceUri || '',
-                                                  opt_rootTagName || '',
-                                                  null);
-  } else if (typeof ActiveXObject != 'undefined') {
+  if (goog.dom.xml.ACTIVEX_SUPPORT) {
     var doc = goog.dom.xml.createMsXmlDocument_();
     if (doc) {
       if (opt_rootTagName) {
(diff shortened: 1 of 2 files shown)


joeltine:
Thanks! Looks like some lint issues in the PR.

vincenta:
Yes, sorry about that, one line is too long... :/
Should I submit a new pull request ?

joeltine:
Nope, just push your changes to your branch and the tests will be run again.

joeltine:
So I ran the tests on this, and it fails in IE11. It works in IE9 and IE8 though. Here's the test case I'm running: 

````
function testSelectSingleNodeInIe() {                                            
  var xml = goog.dom.xml.loadXml('<a><b><c>d</c></b></a>');                      
  var node = xml.firstChild;                                                     
  var bNode = goog.dom.xml.selectSingleNode(node, 'b');                          
  assertNotNull(bNode);                                                          
}     
```

With result

```
ERROR in testSelectSingleNodeInIe
Expected not to be <null>
```

You might as well add the test to this PR as well. You'll need it to repro and we're going to need it anyway.

concavelenz:
http://stackoverflow.com/questions/19638981/window-activexobject-difference-in-ie11

closure/goog/dom/xml.js
```diff
@@ -39,6 +39,33 @@ goog.dom.xml.MAX_ELEMENT_DEPTH = 256; // Same default as MSXML6.
...
+ * Check for ActiveXObject support by the browser.
+ * @return {boolean} true if browser has ActiveXObject support.
+ * @private
+ */
+goog.dom.xml.hasActiveXObjectSupport_ = function() {
+  try {
+    // Due to lot of changes in IE 9, 10 & 11 behaviour, trying to create the
+    // ActiveXOjbect is a lot more reliable than testing for the existance of
+    // window.ActiveXObject
+    new ActiveXObject('MSXML2.DOMDocument');
```
concavelenz: I'm going to get complaints about this as folks like code to run exception clean so they can "pause on throw" in the debuggers.  Also, for compiled code, correct behavior here relies on creating ActiveXObject being detected as having side-effects which as far as I know isn't actually the case.  Also if it isn't eliminated, calling this in global scope means the code will never be eliminated.

Generally, I think using using goog.userAgent.IE will be a lot cleaner, and allow the code to be eliminated when the appropriate "ASSUME" defines are set.  (MS Edge is not considered IE by the user agent code).

vincenta: As ActiveX usage can be disabled using MSIE's security level, it is not reliable to only use goog.userAgent.IE to test for usage of ActiveXObject... In fact, I do think that trying to instantiate the ActiveXObject is the only reliable way of testing for ActiveX support.

For the global scope issue, is this a better (not perfect) solution ?
```javascript
if (goog.userAgent.IE) {
  /**
   * Check for ActiveXObject support by the browser.
   * @return {boolean} true if browser has ActiveXObject support.
   * @private
   */
  goog.dom.xml.hasActiveXObjectSupport_ = function() {
    [...]
  };

  /**
   * True if browser has ActiveXObject support.
   * Possible override if this test become wrong in coming IE versions.
   * @type {boolean}
   */
  goog.dom.xml.ACTIVEX_SUPPORT =
    goog.dom.xml.hasActiveXObjectSupport_();

} else {
  /**
   * True if browser has ActiveXObject support.
   * Possible override if this test become wrong in coming IE versions.
   * @type {boolean}
   */
  goog.dom.xml.ACTIVEX_SUPPORT = false;
}
```

concavelenz: I would find this acceptable, if you write it so property declarations weren't not conditional on the user agent:

/** ... */
goog.dom.xml.ACTIVEX_SUPPORT = goog.userAgent.IE && goog.dom.xml.hasActiveXObjectSupport_();

This would allow both the code removal and avoid the exception on everything but IE.

joeltine:
Can you squash your commits please? https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

closure/goog/dom/xml.js
```diff
@@ -39,6 +40,39 @@ goog.dom.xml.MAX_ELEMENT_DEPTH = 256; // Same default as MSXML6.
...
+  }
+};
+
+
+/**
+ * True if browser has ActiveXObject support.
+ * Possible override if this test become wrong in coming IE versions.
+ * @type {boolean}
+ */
+goog.dom.xml.ACTIVEX_SUPPORT = goog.userAgent.IE &&
```
joeltine: Can you move this up with the other constants?

vincenta: goog.dom.xml.ACTIVEX_SUPPORT value depends on the return value of the previously declared function...

Should I declare goog.dom.xml.ACTIVEX_SUPPORT before the function and set its value later ?

joeltine: Oops. NVM. Didn't notice.

vincenta:
Commits are squashed.

joeltine:
Unfortunately had to roll this back internally. Will look into a fix on Monday.

joeltine:
This fails for all IE versions now that it's returning `ActiveXObject` for the document. The root cause is someone trying to, ultimately, do this:

```
var doc = new ActiveXObject('MSXML2.DOMDocument');
doc.appendChild(document.createElement('body')); // Throws "Type mismatch"
```

The actual Closure code is:

```
var dummyDoc = goog.dom.xml.createDocument('html');
goog.dom.appendChild(dummyDoc.firstChild, goog.dom.createDom('body'));
```

PR #544, opened 2015-09-02
fixes range's getStartPosition/getEndPosition

joeltine:
Branched from https://github.com/google/closure-library/pull/290


joeltine:
This passed a global presubmit. LGTM.

PR #545, opened 2015-09-02
Get correct parameter for entity name

lewistg:
I signed it!

### DIFF
--- a/third_party/closure/goog/caja/string/html/htmlparser.js
+++ b/third_party/closure/goog/caja/string/html/htmlparser.js
@@ -518,8 +518,10 @@ goog.string.html.HtmlParser.prototype.stripNULs_ = function(s) {
  */
 goog.string.html.HtmlParser.prototype.unescapeEntities_ = function(s) {
   return s.replace(
-      goog.string.html.HtmlParser.ENTITY_RE_,
-      goog.bind(this.lookupEntity_, this));
+      goog.string.html.HtmlParser.ENTITY_RE_, goog.bind(
+          function(fullEntity, name) {
+               return this.lookupEntity_(name);
+          }, this));
 };
 
 


pauldraper:
It looks one of the lines is too long.

You can run

```sh
scripts/ci/install_closure_deps.sh
scripts/ci/lint_pull_request.sh
```

It will check the files that have changed from master.

joeltine:
What problem is this PR trying to solve?

joeltine:
Also, we have no CLA for lewistg

lewistg:
@joeltine The issue I experienced related to the regular expression used [here] (https://github.com/google/closure-library/pull/545/files#diff-4d487608ccab672d5ee2a1bcef130b44L521). [It contains a capturing group] (https://github.com/google/closure-library/pull/545/files#diff-4d487608ccab672d5ee2a1bcef130b44L274), that will be passed into the callback [here] (https://github.com/google/closure-library/pull/545/files#diff-4d487608ccab672d5ee2a1bcef130b44L522) as the *second* parameter. The first is the full match.

Here is a little example that exposes the bug:

`goog.string.html.htmlSanitize('', function(url){return url;});`

results in

``.

The ampersands were deleted because of the bug.

lewistg:
Also, I signed the CLA, but I am not sure how I can verify my consent. I will try again maybe.

joeltine:
Huh, seems crazy such a blatant bug would exist this long, but I'm not too surprised as there's no tests.

third_party/closure/goog/caja/string/html/htmlparser.js
```diff
@@ -474,11 +474,14 @@ goog.string.html.HtmlParser.prototype.parse = function(handler, htmlText) {
...
  * Decodes an HTML entity.
  *
+ * @param {string} fullEntity The full entity including the leading
+ * '&', the trailing ';', and the content in between.
  * @param {string} name The content between the '&' and the ';'.
  * @return {string} A single unicode code-point as a string.
  * @private
  */
-goog.string.html.HtmlParser.prototype.lookupEntity_ = function(name) {
+goog.string.html.HtmlParser.prototype.lookupEntity_ = function(fullEntity,
```
joeltine: It's a bit odd to have a documented method with a required param that does nothing. I think I'd prefer to leave this with just `name`, and just have the replace in  `unescapeEntities_` call `lookupEntity_` correctly with the right thing.

third_party/closure/goog/caja/string/html/htmlparser.js
```diff
@@ -518,8 +518,10 @@ goog.string.html.HtmlParser.prototype.stripNULs_ = function(s) {
  */
 goog.string.html.HtmlParser.prototype.unescapeEntities_ = function(s) {
   return s.replace(
-      goog.string.html.HtmlParser.ENTITY_RE_,
-      goog.bind(this.lookupEntity_, this));
+      goog.string.html.HtmlParser.ENTITY_RE_, goog.bind(
```
lewistg: @joeltine I force pushed, so it seems your original comment was deleted. But, I agree about the unused parameter.

Issue #549, opened 2015-09-07
goog.string.hashCode is a little slow.


weiwl:
The implementation is using `mod` operation. 

```javascript
goog.string.hashCode = function(str) {
  var result = 0;
  for (var i = 0; i < str.length; ++i) {
    result = 31 * result + str.charCodeAt(i);
    // Normalize to 4 byte range, 0 ... 2^32.
    result %= goog.string.HASHCODE_MAX_;
  }
  return result;
};
```

But instead of using mod (%), we can replace it with bitwise OR and addition (+).
It can short the consuming time to 1/3 of original for almost all modern browsers.

```javascript
var bitwise_or_hashCode = function(str) {
  var result = 0;
  for (var i = 0; i < str.length; ++i) {
    result = 31 * result + str.charCodeAt(i);
    result |= 0; // will convert to a 32-bit integer, but the 32nd bit will be used as the sign.
    if (result < 0) {
      result += 4294967296;
    }
  }
  return result;
};
```

Please refer to http://jsperf.com/goog-string-hashcode-vs-improved for performance comparison.

The code below should also work, and be a little faster.
```javascript
var bitwise_or_hashCode = function(str) {
  var result = 0;
  for (var i = 0; i < str.length; ++i) {
    result = 31 * result + str.charCodeAt(i);
    result |= 0; // only keep 32 bits
  }
  if (result < 0) {
    result += 4294967296;
  }
  return result;
};
```

joeltine:
For this to be considered, it'd need to be a PR.

PR #553, opened 2015-09-09
refs #549 improve performance of goog.string.hashCode.

weiwl:
Improve performance of goog.string.hashCode.

### DIFF
--- a/closure/goog/string/string.js
+++ b/closure/goog/string/string.js
@@ -1266,14 +1266,6 @@ goog.string.compareElements_ = function(left, right) {
 
 
 /**
- * Maximum value of #goog.string.hashCode, exclusive. 2^32.
- * @type {number}
- * @private
- */
-goog.string.HASHCODE_MAX_ = 0x100000000;
-
-
-/**
  * String hash function similar to java.lang.String.hashCode().
  * The hash code for a string is computed as
  * s[0] * 31 ^ (n - 1) + s[1] * 31 ^ (n - 2) + ... + s[n - 1],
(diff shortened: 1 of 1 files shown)


weiwl:
We can prove that the two functions below are equivalent.

```javascript
function(str) {
  var result = 0;
  for (var i = 0; i < str.length; ++i) {
    result = 31 * result + str.charCodeAt(i);
    result |= 0;
    if (result < 0) {
      result += goog.string.HASHCODE_MAX_;
    }
  }
  return result;
};
```
```javascript
function(str) {
  var result = 0;
  for (var i = 0; i < str.length; ++i) {
    result = 31 * result + str.charCodeAt(i);
    result |= 0;
  }
  if (result < 0) {
    result += goog.string.HASHCODE_MAX_;
  }
  return result;
};
```
But for chrome browser, the second function is slower than the first one.
See http://jsperf.com/goog-string-hashcode-vs-improved/2 for reference.

closure/goog/string/string.js
```diff
@@ -1289,7 +1289,10 @@ goog.string.hashCode = function(str) {
   for (var i = 0; i < str.length; ++i) {
     result = 31 * result + str.charCodeAt(i);
     // Normalize to 4 byte range, 0 ... 2^32.
-    result %= goog.string.HASHCODE_MAX_;
+    result |= 0;
```
concavelenz: I think simply "result = (...) >>> 0" ?

weiwl: Yes, you are right. It is more simple. Thanks.

PR #558, opened 2015-09-15
Change default country for 'pt' locale

samaxes:
The default country for the locale 'pt' should be Portugal instead of Brazil.

### DIFF
--- a/closure/goog/i18n/compactnumberformatsymbols.js
+++ b/closure/goog/i18n/compactnumberformatsymbols.js
@@ -7123,6 +7123,89 @@ goog.i18n.CompactNumberFormatSymbols_pt = {
       'other': '000\u00A0mil'
     },
     '1000000': {
+      'other': '0\u00A0M'
+    },
+    '10000000': {
+      'other': '00\u00A0M'
+    },
+    '100000000': {
+      'other': '000\u00A0M'
+    },
+    '1000000000': {
+      'other': '0\u00A0MM'
+    },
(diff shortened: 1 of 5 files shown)


closure/goog/i18n/compactnumberformatsymbols.js
```diff
@@ -221,8 +221,8 @@ goog.provide('goog.i18n.CompactNumberFormatSymbols_pa_Guru_IN');
 goog.provide('goog.i18n.CompactNumberFormatSymbols_pl');
 goog.provide('goog.i18n.CompactNumberFormatSymbols_pl_PL');
 goog.provide('goog.i18n.CompactNumberFormatSymbols_pt');
-goog.provide('goog.i18n.CompactNumberFormatSymbols_pt_BR');
 goog.provide('goog.i18n.CompactNumberFormatSymbols_pt_PT');
+goog.provide('goog.i18n.CompactNumberFormatSymbols_pt_BR');
```
shicks: These should be alphabetized

fabalbon:
Hi Samuel,

Closure's locale data is generated from CLDR (the industry standard locale data repository - http://cldr.unicode.org/) which uses pt_BR as the default locale for pt. We have to keep Closure consistent with CLDR therefore to change this you should submit a request directly to CLDR (http://cldr.unicode.org/index/bug-reports) and it will be automatically updated into Closure when CLDR is updated.

Regards!

PR #560, opened 2015-09-17
Add standard browsers to CI tests

joeltine:


shicks:
Looks reasonable to me.

joeltine:
Note: unsurprisingly the tests fail in all IE version for one reason or another. For now, only enabling FF+Chrome.

PR #561, opened 2015-09-17
Add goog.array.partition

igrep:
Like Underscore.js's partition method: http://underscorejs.org/#partition .

I'm not sure the policy how you decide which functions to add to `goog.array`.
So feel free to close with the reason.

### DIFF
--- a/closure/goog/array/array.js
+++ b/closure/goog/array/array.js
@@ -605,7 +605,7 @@ goog.array.findIndexRight = function(arr, f, opt_obj) {
  *     takes 3 arguments (the element, the index and the array) and should
  *     return a boolean.
  * @param {S=} opt_obj An optional "this" context for the function.
- * @return {Array<Array<T>, Array<T>>} An array of 2 elements: values of
+ * @return {Array<Array<T>>} An array of 2 elements: values of
  *       arr which satisfies the condition and ones which does not.
  * @template T,S
  */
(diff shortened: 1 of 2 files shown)


closure/goog/array/array.js
```diff
@@ -596,6 +596,34 @@ goog.array.findIndexRight = function(arr, f, opt_obj) {
...
+ * Split array into two arrays: one that satisfies the given condition,
+ * and one that does not satisfy the condition.
+ * @param {Array<T>|goog.array.ArrayLike} arr Array or array
+ *     like object over which to iterate.
+ * @param {?function(this:S, T, number, ?) : boolean} f The function to call
+ *     for every element. This function
+ *     takes 3 arguments (the element, the index and the array) and should
+ *     return a boolean.
+ * @param {S=} opt_obj An optional "this" context for the function.
+ * @return {Array<Array<T>, Array<T>>} An array of 2 elements: values of
```
concavelenz: we dont have tuples: Array<Array<T>> is correct

shicks:
`goog.array.bucket` should be able to do what you need.  You'll get an object `{'true': [...], 'false': [...]}` but otherwise it's the same thing (and possibly better since it's clear which element is which).

igrep:
I tried `goog.array.bucket` in my case, looks fine though a bit verbose.

```javascript
    ( // <- parenthesis required when using desctructuring assignment like below.
      { 'true': this.dones, 'false': this.notYets } = goog.array.bucket(schedules, (
        /** KeepMeContributing.Worker.TimeOfDay */ timeOfDay
      ) => {
        return timeOfDay.millisecsAfter(now) <= 0;
      })
    );
```

Thanks for advice! :+1: 
I'll close.

PR #562, opened 2015-09-18
Fix IE tests on sauce labs

joeltine:


joeltine:
Ignore this PR for now. Just trying to get IE passing on Sauce. (It's having all sorts of trouble)

Issue #542, opened 2015-09-01
goog.userAgent.flash version detection fails on IE11 for flash player version 11


smudgeface:
See this similar issue with swfobject:
https://github.com/swfobject/swfobject/issues/30

This fiddle shows the issue (when run on IE11 with flash player version 11.X)
http://jsfiddle.net/noyobo/hEPb7/11/show/

Basically, IE11 has introduced support for navigator.plugins, but older versions of Flash player do not play nicely. The fallback mechanism (using ActiveXObject) should be preferred if the UA is IE, rather than simply relying on the existence of navigator.plugins.

joeltine:
PRs welcome

shicks:
This issue is being closed as a preparation step before archiving the repository.  See the [README](https://github.com/google/closure-library) for more detail.

PR #566, opened 2015-09-23
Addressing issue #542

plu83:
A return now occurs in the flash init function when the detected Flash
Player version is assigned a value, otherwise we attempt the other means
of retrieving the version until no further options exist.

### DIFF
--- a/closure/goog/useragent/flash.js
+++ b/closure/goog/useragent/flash.js
@@ -60,51 +60,61 @@ goog.userAgent.flash.init_ = function() {
       if (plugin.description) {
         goog.userAgent.flash.detectedFlashVersion_ =
             goog.userAgent.flash.getVersion_(plugin.description);
+        return;
       }
     }
 
     if (navigator.plugins['Shockwave Flash 2.0']) {
       goog.userAgent.flash.detectedFlash_ = true;
       goog.userAgent.flash.detectedFlashVersion_ = '2.0.0.11';
+      return;
     }
+  }
 
-  } else if (navigator.mimeTypes && navigator.mimeTypes.length) {
+  if (navigator.mimeTypes && navigator.mimeTypes.length) {
     var mimeType = navigator.mimeTypes['application/x-shockwave-flash'];
     goog.userAgent.flash.detectedFlash_ = mimeType && mimeType.enabledPlugin;
     if (goog.userAgent.flash.detectedFlash_) {
       goog.userAgent.flash.detectedFlashVersion_ =
           goog.userAgent.flash.getVersion_(mimeType.enabledPlugin.description);
+      return;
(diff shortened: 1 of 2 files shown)


closure/goog/useragent/flash.js
```diff
@@ -60,49 +60,52 @@ goog.userAgent.flash.init_ = function() {
...
-  } else {
+  /** @preserveTry */
+  try {
+    // Try 7 first, since we know we can use GetVariable with it
+    var ax = new ActiveXObject('ShockwaveFlash.ShockwaveFlash.7');
+    goog.userAgent.flash.detectedFlash_ = true;
+    goog.userAgent.flash.detectedFlashVersion_ =
+        goog.userAgent.flash.getVersion_(ax.GetVariable('$version'));
+  } catch (e) {
+    // Try 6 next, some versions are known to crash with GetVariable calls
```
shicks: It would be more consistent with the other changes you're making to add a return at the end of the try blocks and then have the catch simply fall through.

shicks:
Please squash these commits into a single one we can pull.  Thanks!

plu83:
Thanks for the feedback, @shicks. I've added the additional return statements and squashed the commits into 26076c6.

closure/goog/useragent/flash.js
```diff
@@ -60,49 +60,55 @@ goog.userAgent.flash.init_ = function() {
...
-  } else {
+  /** @preserveTry */
+  try {
+    // Try 7 first, since we know we can use GetVariable with it
+    var ax = new ActiveXObject('ShockwaveFlash.ShockwaveFlash.7');
+    goog.userAgent.flash.detectedFlash_ = true;
+    goog.userAgent.flash.detectedFlashVersion_ =
+        goog.userAgent.flash.getVersion_(ax.GetVariable('$version'));
+    return;
+  } catch (e) {
```
shicks: This should be simply

    } catch (e) { /* fall-through */ }

Since you're not using the exceptions, there's no need to do the work inside the `catch` block, and introduce an extra level of indentation and require new names `e2`, and `e3`.

shicks:
It doesn't look like you were able to squash these *on top of* `master`, but rather *merged* `master` into it, which isn't the same thing.  The trick is that when you push back to your own github repository, you need to use `git push -f github master` (or whatever your remote is called).  Otherwise, git will complain that your push would rewrite history and suggest you pull and merge (which is what you did) first.  But in this case, rewriting your fork's history is exactly what you want to do.

plu83:
Thanks for the information, @shicks and sorry for my newbie-ness. As you might have guessed, this was my first squashing. It's a great idea, particularly with large repositories like Closure Library.

PR #568, opened 2015-09-27
Refactor duplicated function expressions for extracting hex and alpha colors

kencheung:


closure/goog/color/alpha.js
```diff
@@ -68,21 +68,26 @@ goog.color.alpha.hexToRgbaStyle = function(hexColor) {
 };
 
 
+function extractColor(startIndex, endIndex) {
```
concavelenz: - This isn't a goog.module, you don't want to introduce a global name here.
- This is missing proper jsdoc and type declarations
- It would be better to call a private method with parameters, rather than to return a function expression.  The former is more regular and more likely to be optimized by VMs and the Closure Compiler.

kencheung:
I changed to private method and added jsdoc as suggested. I don't understand why the build fails on only firefox on linux, even though the code does not seem to be browser dependent.

closure/goog/color/alpha.js
kencheung: I changed to private method and added jsdoc as suggested. I don't understand why the build fails on only firefox on linux, even though the code does not seem to be browser dependent.

closure/goog/color/alpha.js
```diff
@@ -69,16 +69,18 @@ goog.color.alpha.hexToRgbaStyle = function(hexColor) {
...
 /**
- * Gets the hex color part of an alpha hex color. For example, from '#abcdef55'
- * return '#abcdef'.
+ * Gets the hex or alpha color part of an alpha hex color
  * @param {string} colorWithAlpha The alpha hex color to get the hex color from.
+ * @param {number} startIndex The start index of the alpha hex color to strip
+   from.
+ * @param {number} endIndex The end index of the alpha hex color to strip until.
  * @return {string} The hex color where the alpha part has been stripped off.
  */
```
concavelenz: @private please.

closure/goog/color/alpha.js
```diff
@@ -69,16 +69,18 @@ goog.color.alpha.hexToRgbaStyle = function(hexColor) {
...
-goog.color.alpha.extractHexColor = function(colorWithAlpha) {
+goog.color.alpha.extractColor = function(colorWithAlpha, startIndex, endIndex) {
```
concavelenz: private methods should be named with a trailing underscore:
goog.color.alpha.extractColor_

thanks.

kencheung: Revised as suggested. Thank you.

shicks:
Looks good.  Can you please squash this down to a single commit so we can pull it?

kencheung:
I have squashed my commits into a single one.

I notice that many of my builds fail on particular browsers. I am aware that the compatibility of FF41 may be one of the causes, but sometimes they fail also on particular versions of IE. What will be possible causes for my failed builds?

As I can test my code on only a single platform and I do not have access to the build reports in SauceLabs, I am interested in how certain commits or upgrade of certain browsers lead to build failures.

joeltine:
@kencheung I took a look at the test failures in IE, and they're not related to your PR. Don't worry about it.

shicks:
I still see "9 commits" up at the top.  Make sure that after you rebase on top of our master branch, you use `git push -f` to push it back to your fork.  Otherwise the push will fail and suggest you pull and merge, which defeats the whole purpose of squashing.  Try `git log --graph --decorate --color --all` to see more clearly what I mean.

kencheung:
I think it works this time. Thank you very much for your help.

PR #569, opened 2015-09-28
Changed return type of goog.object.get

ducklord:
Hey,
This is a change to something very central, so if you do not want this change I do understand. But is it not true that when the optional value is set, the return value will always be this value, and in case the default return value is not set, then template value R will be undefined as return type? So the return type will be V|undefined, when opt_val is not given as parameter?

### DIFF
--- a/closure/goog/object/object.js
+++ b/closure/goog/object/object.js
@@ -396,7 +396,7 @@ goog.object.add = function(obj, key, val) {
  * @param {string} key The key for which to get the value.
  * @param {R=} opt_val The value to return if no item is found for the given
  *     key (default is undefined).
- * @return {V|R|undefined} The value for the given key.
+ * @return {V|R} The value for the given key.
  * @template K,V,R
  */
 goog.object.get = function(obj, key, opt_val) {


shicks:
This doesn't actually work quite how you'd expect.  I opened https://github.com/google/closure-compiler/issues/1176 to address the underlying problem.

shicks:
Even with the compiler bug, this is still potentially helpful.  But it also has the potential to break a lot of tests.  I'm running it through our internal testing tools to see what it does before pulling.  Thanks for finding this!

shicks:
Looks like nothing's broken as a result.  @concavelenz what do you think?

joeltine:
FYI: Rebasing on the latest master should fix the test failure on Travis. Fixed in https://github.com/google/closure-library/commit/c82f7daf8da79ec0e9164478728b634d5581a8aa.

ducklord:
What the ..... happened there. Tried to make a rebase, Something went very wrong there.

ducklord:
The diff looks clean, but after the rebase, for some reason the unit test fails at various places that is not touched by the patch. Like a key variable that is used in for each loops, but the key value is never used.

joeltine:
@ducklord You can ignore those lint warnings. They're expected.

concavelenz:
This change can't land until such a time as the compiler can describe this.

shicks:
We're working on a change that will finally make this possible by assuming unmatched typevars are bottom.  But it's blocked on a large depot change and is currently stalled.

shicks:
Interestingly, TypeScript can't infer this correctly, either.  At least, not without overloads, which are not likely to ever land in Closure Compiler.

That said, this function is still somewhat problematic.  There are two or three distinct use cases that are being conflated into one: (1) allows indexing an Object<K, V> with a string, (2) allows setting a default, (3) allows null-safe access.  The first two are reasonable to call without the third argument, but (1) is a misfeature - an explicit cast would be better - and (2) is soon to be obsolete once [optional chaining](https://github.com/tc39/proposal-optional-chaining) lands.  Given that, it makes sense to just make the third argument required.

brad4d:
Internal Google issue http://b/142066247 created

brad4d:
Unfortunately it appears to be infeasible to accept this, as much as we would like to do so.

PR #571, opened 2015-09-29
Refactor duplicated function expressions for color factory

kencheung:


shicks:
This is adding a lot of code - a whole new function that takes 6 parameters - and I'm not convinced it's pulling its weight, since most of what it's "factoring out" is simply passing parameters through to a factory method.

I could get behind factoring out the _rest_ of the methods, though - if you wanted to define something that took a button and a starting color and attached what needed to be attached.  Then you'd have (abbreviated)

    ....fontColor = function(...) {
      var button = ....makeColorMenuButton(...);
      ....decorateColorMenuButton_(button, '#000');
      return button;
    }

or even

    return ....decorateColorMenuButton_(....makeColorMenuButton(...), '#000');

PR #583, opened 2015-10-06
Use native Array.prototype.findIndex function

fredj:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

### DIFF
--- a/closure/goog/array/array.js
+++ b/closure/goog/array/array.js
@@ -529,16 +529,26 @@ goog.array.find = function(arr, f, opt_obj) {
  *     or -1 if no element is found.
  * @template T,S
  */
-goog.array.findIndex = function(arr, f, opt_obj) {
-  var l = arr.length;  // must be fixed during loop... see docs
-  var arr2 = goog.isString(arr) ? arr.split('') : arr;
-  for (var i = 0; i < l; i++) {
-    if (i in arr2 && f.call(opt_obj, arr2[i], i, arr)) {
-      return i;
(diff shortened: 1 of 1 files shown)


closure/goog/array/array.js
```diff
@@ -501,7 +501,7 @@ goog.array.count = function(arr, f, opt_obj) {
  * return that element.
  * @param {Array<T>|goog.array.ArrayLike} arr Array or array
  *     like object over which to iterate.
- * @param {?function(this:S, T, number, ?) : boolean} f The function to call
+ * @param {!function(this:S, T, number, ?) : boolean} f The function to call
```
joeltine: While correct, these type changes have the potential to break a bunch of ppl.

joeltine:
There are some compilation errors in the travis report.

fredj:
This pull request depends on https://github.com/google/closure-compiler/pull/1190

concavelenz:
Generally, I would rather keep adding these unless there is some inherit performance advantage to the native methods (which there generally is not as the VMs just write these methods in JS).

joeltine:
@concavelenz did you mean "rather not"?

concavelenz:
sorry, yes "rather not".   Is there a clear performance advantage?

joeltine:
@fredj Got a jsperf?

fredj:
I don't need this anymore, closing

Issue #570, opened 2015-09-28
Refine type information on createElement()'s returns


psam44:
This is a continuation on https://github.com/google/closure-library/pull/503, related to https://github.com/google/closure-compiler/issues/1054.

In base.js:
'Element' is too broad as type and prevents disambiguation on property names.

A fix is:
line 1130
change

```
var scriptEl = doc.createElement('script');
```

to

```
var scriptEl = /** @type {HTMLScriptElement} */ (doc.createElement('script'));
```

line 1922
change

```
  var scriptElt = doc.createElement('SCRIPT');
```

to

```
  var scriptElt = /** @type {HTMLScriptElement} */ (doc.createElement('SCRIPT'));
```

Note: A cast is used because a simple type declaration leaves a `WARNING - initializing variable` when compiled with `--warning_level=VERBOSE`.

shicks:
Would you be up for putting together a pull request for this?

psam44:
I don't have the repository to make a PR. The fix is only two tiny copy-paste.

shicks:
You can use GitHub's button to fork this repository.  It may only be a few small changes, but a commit needs to be made, and merged, etc.  Since it's a low-priority issue, we would prefer if you could make that change and then we can just pull it in after Travis tests it.

psam44:
PR done. There is a Travis fail, but it isn't on my changes.

bowbahdoe:
This issue should be closed, a PR was merged that fixed it as mentioned in #584

PR #584, opened 2015-10-09
Refine type returns of createElement()

psam44:
For https://github.com/google/closure-library/issues/570

### DIFF
--- a/closure/goog/base.js
+++ b/closure/goog/base.js
@@ -1129,7 +1129,8 @@ if (goog.DEPENDENCIES_ENABLED) {
   goog.appendScriptSrcNode_ = function(src) {
     /** @type {Document} */
     var doc = goog.global.document;
-    var scriptEl = doc.createElement('script');
+    var scriptEl = /** @type {HTMLScriptElement} */
+        (doc.createElement('script'));
     scriptEl.type = 'text/javascript';
     scriptEl.src = src;
     scriptEl.defer = false;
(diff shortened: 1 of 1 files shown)


closure/goog/base.js
```diff
@@ -1129,7 +1129,8 @@ if (goog.DEPENDENCIES_ENABLED) {
   goog.appendScriptSrcNode_ = function(src) {
     /** @type {Document} */
     var doc = goog.global.document;
-    var scriptEl = doc.createElement('script');
+    var scriptEl = /** @type {HTMLScriptElement} */
```
shicks: Would it make sense to mark these as {!HTMLScriptElement} ?

shicks:
I'm not sure why the CLA check still fails for you, since you did respond in the last PR.  Maybe it's having trouble linking the address you signed with to your github username?

closure/goog/base.js
psam44: You're probably right, because of {!Element} of Document.prototype.createElement.

psam44:
CLA: yes, my primary google email and my github email are different.
Added the github account name, to see if it helps.

closure/goog/base.js
shicks: Okay, please make that change and then we can pull this.

concavelenz:
I'll pull this in manually.  This revealed an issue with the conformance framework, so it needed some work to land.  Should go out today.

joeltine:
Merged in https://github.com/google/closure-library/commit/54b3bebb1698a0ad00ce3dce6a9850e381c7d521#diff-c359099ba4bc09bc2864536c49356d52

PR #586, opened 2015-10-13
UTF-8 encode output

filipecatraia:
Hey,

I use Closure for a project, and compilation breaks when using certain Unicode characters. For example:

```
UnicodeEncodeError: ascii codec can't encode character u\xd7 ...
```

UTF-8 encoding the `out.write` outputs solves this issue for us. I'm not sure how valuable it is for the Closure project, but it could get other people out of trouble :)


nanaze:
lgtm

filipecatraia:
I appreciate the chance to contribute ;)

PR #588, opened 2015-10-15
Fix all tests and remove localhost like one variable

PabloSuarez:
Fix all tests and remove localhost like one variable
over smoke


joeltine:
I can't look at this until a CLA is signed.

PR #593, opened 2015-10-22
define goog.TESTING flag

myphysicslab:
Define goog.TESTING flag which indicates that this is a non-production
build for testing purposes. Use this flag in goog.async.run in addition to
goog.DEBUG.  This allows unit tests that use goog.testing.MockClock to
run without requiring goog.DEBUG to be on.

### DIFF
--- a/closure/goog/async/run.js
+++ b/closure/goog/async/run.js
@@ -102,7 +102,8 @@ goog.async.run.workQueue_ = new goog.async.WorkQueue();
 
 if (goog.DEBUG || goog.TESTING) {
   /**
-   * Reset the work queue. Only available in TESTING mode.
+   * Reset the work queue. Only available for tests in debug mode
+   * or testing mode.
    */
   goog.async.run.resetQueue = function() {
     goog.async.run.workQueueScheduled_ = false;
--- a/closure/goog/async/run.js
+++ b/closure/goog/async/run.js
@@ -100,7 +100,7 @@ goog.async.run.workQueueScheduled_ = false;
 goog.async.run.workQueue_ = new goog.async.WorkQueue();
 
 
-if (goog.TESTING) {
+if (goog.DEBUG || goog.TESTING) {
   /**
    * Reset the work queue. Only available in TESTING mode.
    */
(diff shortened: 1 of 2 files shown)


myphysicslab:
After getting a recent version of closure-library, my unit tests were failing because I run them with goog.DEBUG off.  Turning on goog.DEBUG causes lots of unwanted debug code to write to the console during my unit tests.

The failure message was something like `goog.async.run.resetQueue is undefined` which happened while trying to uninstall a MockClock.

myphysicslab:
Travis is reporting a lint problem on `goog.loadModuleFromSource_` in `base.js`, but I didn't change anything in or near that function.

```
----- FILE  :  /home/travis/build/google/closure-library/closure/goog/base.js -----
Line 1123, E:0251: Invalid suppression type: newCheckTypes
Line 1124, E:0217: Missing @return JsDoc in function with non-trivial return
Found 2 errors, including 0 new errors, in 1 file (1 file OK).
```

This is the function in `goog.base` where the lint errors are being reported:

```javascript
  /**
   * @private @const {function(string):?}
   *
   * The new type inference warns because this function has no formal
   * parameters, but its jsdoc says that it takes one argument.
   * (The argument is used via arguments[0], but NTI does not detect this.)
   * @suppress {newCheckTypes}
   */
  goog.loadModuleFromSource_ = function() {
    // NOTE: we avoid declaring parameters or local variables here to avoid
    // masking globals or leaking values into the module definition.
    'use strict';
    var exports = {};
    eval(arguments[0]);
    return exports;
  };
```

joeltine:
You can ignore the lint. base.js has some known, currently unfixable, lint issues.

This is too specific a use case for Closure though. DEBUG is specifically for non-production modes (including testing, as observed in goog.async.run.resetQueue). You typically want verbose logs in this mode. 

If you want to quiet logs, I'd recommend looking into something like `goog.debug.logger` which allows you to control the log level programmatically: https://github.com/google/closure-library/blob/master/closure/goog/debug/logger.js#L157.  You could even create your own goog.defines that set these levels for you.

concavelenz:
What specific messages did you want to get rid of?  If some code is
particularly noisy we can look at removing some messages, turning them
down, etc.

myphysicslab:
I was mainly concerned with messages in my own code, so using `goog.debug.logger` or using `goog.defines` to make a custom symbol will work OK.

PR #594, opened 2015-10-29
Refactor queue remove to use goog.array.removeLast()

Anish2:


closure/goog/array/array.js
```diff
@@ -700,6 +700,24 @@ goog.array.remove = function(arr, obj) {
...
+ * Removes the last occurrence of a particular value from an array.
+ * @param {!Array<T>|!goog.array.ArrayLike} arr Array from which to remove
+ *     value.
+ * @param {T} obj Object to remove.
+ * @return {boolean} True if an element was removed.
+ * @template T
+ */
+goog.array.removeLast = function(arr, obj) {
+  var i = goog.array.lastIndexOf(arr, obj);
+  var rv;
```
shicks: Might as well just set rv = (i >= 0) here, where it's more straightforward.  Could also use a better name (like isPresent).  Alternatively, don't even bother setting the variable.

    if (i >= 0) {
      // remove
      return true;
    }
    return false;

closure/goog/array/array.js
```diff
@@ -1417,7 +1435,8 @@ goog.array.bucket = function(array, sorter, opt_obj) {
 goog.array.toObject = function(arr, keyFunc, opt_obj) {
   var ret = {};
   goog.array.forEach(arr, function(element, index) {
-    ret[keyFunc.call(/** @type {?} */ (opt_obj), element, index, arr)] = element;
+    ret[keyFunc.call(/** @type {?} */ (opt_obj), element, index, arr)] =
+        element;
```
shicks: This seems unrelated, though it does look like we had an issue with >80 columns.  Might be more readable to just pull out

    var key = keyFunc.call(...);

rather than wrapping the line.

closure/goog/array/array_test.js
```diff
@@ -701,6 +701,16 @@ function testArrayRemove() {
   assertArrayEquals('remove, remove non existing element', ['a', 'b', 'd'], a);
 }
 
+function testArrayRemoveLast() {
+  var a = ['c', 'a', 'b', 'c', 'd', 'a'];
+  goog.array.removeLast(a, 'c');
+  assertArrayEquals('remove, remove existing element', ['c', 'a', 'b', 'd', 'a'], a);
```
shicks: Please wrap these long lines to 80 chars

shicks:
Thanks for putting this together.  Before we can merge, we're going to need you to squash these to a single commit on top of our `master` branch (note, you'll need to use `git push -f` after rebasing to avoid having to pull and merge).

PR #595, opened 2015-11-02
Changed danish group-seperator

Squazz:
Changed danish group-seperator to follow danish standards for showing timestamps

### DIFF
--- a/closure/goog/i18n/numberformatsymbols.js
+++ b/closure/goog/i18n/numberformatsymbols.js
@@ -716,7 +716,7 @@ goog.i18n.NumberFormatSymbols_cy_GB = goog.i18n.NumberFormatSymbols_cy;
  */
 goog.i18n.NumberFormatSymbols_da = {
   DECIMAL_SEP: ',',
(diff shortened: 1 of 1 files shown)


joeltine:
These files are auto-generated based on [CLDR](http://cldr.unicode.org/). If it's correct there, it's possible we haven't pushed the latest changes. If it isn't, you can file a bug against CLDR.

fabalbon:
As joeltine said, Closure's locale data is generated from CLDR (the industry standard locale data repository - http://cldr.unicode.org/) which uses '.' as the grouping separator for the 'da' locale. We have to keep Closure consistent with CLDR therefore to change this you should submit a request directly to CLDR (http://cldr.unicode.org/index/bug-reports) and it will be automatically updated into Closure when CLDR is updated.

Regards!

Squazz:
@joeltine thanks, it was also wrong there, and I have therefor filed a bug there: http://unicode.org/cldr/trac/ticket/9060
@fabalbon thanks for the link for bug-reports :+1:

PR #599, opened 2015-11-05
Configure windows VM to always disallow cross-origin XHRs

joeltine:
One of our XHR tests tries to hit google.com expecting it to fail due to same origin policy. Sauce actually enables same origin requests by default on Windows, so we need to disable it here.


concavelenz:
LG

joeltine:
Unfortunately, this doesn't work. It was recommended to me by Sauce labs support... the registry command is probably wrong.

PR #602, opened 2015-11-12
use &lt; instead of < in comment

myphysicslab:
In `goog.events.eventid` the comment has sample code which includes text
that looks like an HTML tag.  The text is:

     /** @type {goog.events.EventId<MyEventObj>}

When building documentation with js-dossier, this comment generates
warnings:

     [WARNING][com.github.jsdossier.soy.HtmlSanitizer]
     Discarded tag "myeventobj" in text:

Change to use `&lt;` instead of `<` in that comment to prevent these
warnings.  Additionally, use `<pre>` tags so that the comment appears
as a comment block with the original line breaks (instead of being
run together into a long line of text).  Further, add the closing comment
symbol * / to the code sample (with a space so it doesn't actually close
the comment block).


myphysicslab:
I don't understand the Travis CI errors.  This is only a change to a comment, hard to see how it could change any test results.

joeltine:
Various asynchronous tests are flaky on the Travis+Sauce Labs
infrastructure. You can ignore them.

joeltine:
There are 10 commits in this PR. Can you get it down to 1?

js-dossier supports markdown style code blocks: http://daringfireball.net/projects/markdown/syntax#precode.  Can you get this working with 4 indented spaces instead? It's easier to read that way.

myphysicslab:
I will make a new fork to get rid of those old commits (or is there a better way to do that?) and resubmit the PR.

The `<pre>` tags are necessary otherwise the `&lt;` is not interpreted.  I can indent, but then the result will have extra indentation.  Here are examples in github's markdown, which has similar behavior:


Pre-tag without extra indent
<pre>
/** @type {goog.events.EventId&lt;MyEventObj&gt;} * /
var myEventId = new goog.events.EventId(
    goog.events.getUniqueId(('someEvent'));

// No need to cast or declare here since the compiler knows the
// correct type of 'evt' (MyEventObj).
something.listen(myEventId, function(evt) {});
</pre>

Pre-tag with extra indent

<pre>
    /** @type {goog.events.EventId&lt;MyEventObj&gt;} * /
    var myEventId = new goog.events.EventId(
        goog.events.getUniqueId(('someEvent'));

    // No need to cast or declare here since the compiler knows the
    // correct type of 'evt' (MyEventObj).
    something.listen(myEventId, function(evt) {});
</pre>

No pre-tag, just indent -- note that `&lt;` is not interpreted:

    /** @type {goog.events.EventId&lt;MyEventObj&gt;} * /
    var myEventId = new goog.events.EventId(
        goog.events.getUniqueId(('someEvent'));

    // No need to cast or declare here since the compiler knows the
    // correct type of 'evt' (MyEventObj).
    something.listen(myEventId, function(evt) {});

PR #603, opened 2015-11-14
use markdown code block in comment of eventid

myphysicslab:
In `goog.events.eventid` the comment has sample code which includes text
that looks like an HTML tag.  The text is:

     /** @type {goog.events.EventId<MyEventObj>}

When building documentation with js-dossier, this comment generates
warnings:

     [WARNING][com.github.jsdossier.soy.HtmlSanitizer]
     Discarded tag "myeventobj" in text:

Change to use `&lt;` instead of `<` in that comment to prevent these
warnings. Additionally, use `<pre>` tags so that the `&lt;` is
interpreted correctly, and the line breaks are preserved. Further, add
the closing comment symbol * / to the code sample (with a space so it
doesn't actually close the comment block).

### DIFF
--- a/closure/goog/events/eventid.js
+++ b/closure/goog/events/eventid.js
@@ -18,15 +18,14 @@ goog.provide('goog.events.EventId');
 
 /**
  * A templated class that is used when registering for events. Typical usage:
- * <code>
- *   /** @type {goog.events.EventId<MyEventObj>}
- *   var myEventId = new goog.events.EventId(
+ *
+ *    /** @type {goog.events.EventId<MyEventObj>} *\
+ *    var myEventId = new goog.events.EventId(
  *       goog.events.getUniqueId(('someEvent'));
  *
- *   // No need to cast or declare here since the compiler knows the correct
- *   // type of 'evt' (MyEventObj).
- *   something.listen(myEventId, function(evt) {});
- * </code>
+ *    // No need to cast or declare here since the compiler knows the
+ *    // correct type of 'evt' (MyEventObj).
+ *    something.listen(myEventId, function(evt) {});
  *
  * @param {string} eventId
  * @template T


myphysicslab:
Here is info about *"using markdown style code blocks with 4 indented spaces"* in response to a question in previous PR #602:

The `<pre>` tags are necessary otherwise the `&lt;` is not interpreted.  I've added some indentation, but note the result has extra indentation.  Here are examples in github's markdown, which has similar behavior:


Pre-tag without extra indent
<pre>
/** @type {goog.events.EventId&lt;MyEventObj&gt;} * /
var myEventId = new goog.events.EventId(
    goog.events.getUniqueId(('someEvent'));

// No need to cast or declare here since the compiler knows the
// correct type of 'evt' (MyEventObj).
something.listen(myEventId, function(evt) {});
</pre>

Pre-tag with extra indent

<pre>
    /** @type {goog.events.EventId&lt;MyEventObj&gt;} * /
    var myEventId = new goog.events.EventId(
        goog.events.getUniqueId(('someEvent'));

    // No need to cast or declare here since the compiler knows the
    // correct type of 'evt' (MyEventObj).
    something.listen(myEventId, function(evt) {});
</pre>

No pre-tag, just indent -- note that `&lt;` is not interpreted:

    /** @type {goog.events.EventId&lt;MyEventObj&gt;} * /
    var myEventId = new goog.events.EventId(
        goog.events.getUniqueId(('someEvent'));

    // No need to cast or declare here since the compiler knows the
    // correct type of 'evt' (MyEventObj).
    something.listen(myEventId, function(evt) {});

closure/goog/events/eventid.js
```diff
@@ -18,15 +18,15 @@ goog.provide('goog.events.EventId');
 
 /**
  * A templated class that is used when registering for events. Typical usage:
- * <code>
- *   /** @type {goog.events.EventId<MyEventObj>}
+ * <pre>
+ *   /** @type {goog.events.EventId&lt;MyEventObj&gt;} * /
```
shicks: If we're going to pollute the example with entities, we might as well use 	&#47; for the closing slash.

closure/goog/events/eventid.js
```diff
@@ -18,15 +18,15 @@ goog.provide('goog.events.EventId');
 
 /**
  * A templated class that is used when registering for events. Typical usage:
- * <code>
- *   /** @type {goog.events.EventId<MyEventObj>}
+ * <pre>
```
shicks: Can we just use a markdown code block here (either ``` or four indents)?

myphysicslab: Unfortunately that seems to not work: then the `&gt;` does not show up as `<` in the docs.  See https://github.com/google/closure-library/pull/603#issuecomment-156789878

joeltine: Do you still need the entity encoding if you're using 4 indents? I assumed
the encoding was only to fix the generated API.

myphysicslab:
Good catch, it turns out that the `&lt;` entity is not needed with 4 indents, and I'm not getting the js-dossier errors either.  Here is an example using github's markdown.

    /** @type {goog.events.EventId<MyEventObj>} * /
    var myEventId = new goog.events.EventId(
       goog.events.getUniqueId(('someEvent'));

    // No need to cast or declare here since the compiler knows the
    // correct type of 'evt' (MyEventObj).
    something.listen(myEventId, function(evt) {});

I made a new commit to remove the `<pre>` tag and just use 4 indents.

myphysicslab:
I've squashed to a single commit.  I don't think adding an extra space on each line of the code sample is helpful because you wind up seeing that extra space at start of each line in the documentation.  I did however change the closing comment symbol `*/` to use a back-slash instead `*\`.  I'm not sure if that's what you were suggesting.... it's odd-looking but I think anyone would understand what it is intended.

For comparison here is the comment with and without the extra space, in github's markdown:

with 4 spaces indent

    /** @type {goog.events.EventId<MyEventObj>} *\
    var myEventId = new goog.events.EventId(
       goog.events.getUniqueId(('someEvent'));

    // No need to cast or declare here since the compiler knows the
    // correct type of 'evt' (MyEventObj).
    something.listen(myEventId, function(evt) {});

with 5 spaces indent

     /** @type {goog.events.EventId<MyEventObj>} *\
     var myEventId = new goog.events.EventId(
        goog.events.getUniqueId(('someEvent'));

     // No need to cast or declare here since the compiler knows the
     // correct type of 'evt' (MyEventObj).
     something.listen(myEventId, function(evt) {});

shicks:
Maybe you misunderstood what I meant by "here" - the "goog" should be lined up with "myEventId" since we use +4 indent for line continuations.  I wasn't intending to indent the entire code block extra.  In any case, it's not a big deal.  I'm pulling this internally.  Thanks for doing this!

concavelenz:
merged

PR #604, opened 2015-11-15
Changed name for better compatibility

Immortalin:
Apps using the closure library usually refer to it as `closure-library/../../` etc. However, the name google-closure-library breaks this and there is currently [no way to alias it](https://github.com/npm/npm/issues/2943) within npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "google-closure-library",
+  "name": "closure-library",
   "description": "Google's common JavaScript library",
   "version": "20151015.0.0",
   "repository": {


joeltine:
@ChadKillingsworth won't this cause confusion for ppl who are used to "google-closure-library"?

ChadKillingsworth:
The 'google-' prefix was added to the projects because templates and compiler names were already taken. The correct path here would be to get a scoped npm organization `@google` to prefix all the names with. It's on my list to bring up today at the compiler meeting.

But yes, changing the name would definitely break existing users - so will adding a scope name.

PR #606, opened 2015-11-17
Fix duplicate declarations of originalDomHelper/originalElement

nicks:
The compiler complains about this in new type inference


shicks:
Merged.

PR #611, opened 2015-11-20
Make IeMouseEventSequenceSimulator_ ES5 strict mode compatible.

hochhaus:
Previously, IeMouseEventSequenceSimulator_#handleClick_ wrote
to {Event.type,MouseEvent.button} which are read only properties.
When running in ES5 strict mode, this causes a run time exception.

To avoid writing the read-only properties, create two new MouseEvents
for the simulated MOUSEDOWN and MOUSEUP events.

### DIFF
--- a/closure/goog/ui/control.js
+++ b/closure/goog/ui/control.js
@@ -1483,6 +1483,18 @@ goog.ui.Control.IeMouseEventSequenceSimulator_ = function(control) {
 goog.inherits(goog.ui.Control.IeMouseEventSequenceSimulator_, goog.Disposable);
 
 
+/**
+ * Whether this browser supports synthetic MouseEvents.
+ *
+ * See https://msdn.microsoft.com/library/dn905219(v=vs.85).aspx for details.
+ *
+ * @private
+ * @const
+ */
+goog.ui.Control.IeMouseEventSequenceSimulator_.SYNTHETIC_EVENTS_ =
+    !goog.userAgent.IE || goog.userAgent.isDocumentModeOrHigher(9);
+
+
 /** @private */
 goog.ui.Control.IeMouseEventSequenceSimulator_.prototype.handleMouseDown_ =
     function() {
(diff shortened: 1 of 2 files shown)


hochhaus:
I signed it!

shicks:
Thanks for working on this - I've sent it to one of our UI folks to review it more thoroughly.  In the mean time, could you look into adding a regression test for this?  Thanks!

hochhaus:
@shicks @dferrara19 Absolutely, I will add a regression test. I think that doing so requires running such a test in strict mode (on browsers that support it). That made me start to wonder, if the goal is for closure-library to be strict mode compatible, why not run all unit tests in strict mode?

Would a change running all tests in strict mode be considered or would you prefer an isolated test which only uses  IeMouseEventSequenceSimulator_ in strict mode?

Thanks for the guidance before I spend too much time on the wrong path.

hochhaus:
@shicks @dferrara19 I added a regression test. However, it requires makeLeftMouseEvent_ to be run in strict mode. If you can think of cleaner way to write this test please let me know and I'll update the code.

Thanks for your help.

hochhaus:
I squashed all of the changes into a single commit to ease reviewing/merging this change. At this point, I think it is ready for review. Please let me know if you have any suggestions for improvement and I will implement them promptly.

joeltine:
@dferrara19 do you have concerns with this?

shicks:
I don't really understand the test, so hopefully @dferrara19 can chime in here.  In terms of enabling all tests in strict mode, my guess is it'll break a lot of things.  @joeltine can consider whether/how this could be feasible.

hochhaus:
@shicks Thanks. The current test confirms that an `ACTION` event will be fired as the result of individual click `MouseEvent`. The reason this is necessary is because on a `MouseEvent` the {type,button} properties are read only. However, this is not true of the fake events sent from `goog.testing.events.fireClickSequence()`. Therefore, `testIeMouseEventSequenceSimulator()` completes successfully (even in strict mode). Since `testIeMouseEventSequenceSimulatorStrictMode()` uses a native `MouseEvent` it triggers the ES5 strict mode error.

RE: running all tests in strict mode, I no longer think that is a good idea. As `goog.testing.events.fireClickSequence()` does not fire real `MouseEvents` doing so would miss many issues anyways.

shicks:
Where I'm confused is how the you're running this particular test in strict mode?  I would have expected to see a 'use strict'; in the test function or something, but maybe we're talking about different meanings for strict?

hochhaus:
@shicks Good point. I wanted to put the `'use strict';` in the test function but then strict mode only applies to that function (not the functions called by it). Therefore, you'll find the `'use strict';'` in `goog.ui.Control.IeMouseEventSequenceSimulator_.makeLeftMouseEvent_`. This is ugly because that is not strictly testing code but I did not see another way to get that function to run in strict mode.

shicks:
I see.  Sorry for dragging this out so much - I'm a little hesitant to make such a big change to goog.ui, and would really like some input from @dferrara19 who has more experience with this codebase than either Joel or myself do.  In any case, I'll run a test to see if this breaks any internal tests.  If not, that will give me at least a bit more confidence to proceed.

Meanwhile, I'm a bit curious - if we needed to jump through such specific hoops to get the test to run in strict mode, then how is it that this bug came up in production in the first place?  My assumption is that you're compiling your code and outputting it with a strict mode wrapper, but would be interested to hear more specifically how it came up.  Thanks!

hochhaus:
@shicks No worries, thanks for helping to move this forward.

In prod we run our entire application in strict mode ('use strict'; at the file level for each compiled module). As all functions are running in strict mode we ran into this problem. Per John Lenz on CLD, I think other Google apps are likely doing the same (but just not using goog.ui). Is doing this intended to be supported by closure-library?

shicks:
@hochhaus Thanks for the background.  Yes, I would say that this is intended use, and so in principal we should support strict mode here.  It's just a question of doing so in the "legacy" goog.ui library.

The internal test seemed to pass okay.  So I'll pull it and hope for the best.

hochhaus:
@shicks Thanks for your work. I understand the pain of dealing with a change of this size to legacy code.

shicks:
Merged in 546bfccde8b9b9c78c43cc028a8cb05d6cf10fe3

hochhaus:
@shicks Thanks!

PR #612, opened 2015-11-20
change args type to ArrayLike

underbyte:
`args_` assignment on line 123 throws an annotation error since `arguments` is an `ArrayLike`.


shicks:
Looks good, thanks for making this change.

concavelenz:
https://github.com/google/closure-library/commit/32ff833db42bf13ee0dc1c6fe961fa90bdb65af1

PR #616, opened 2015-11-26
using yyyy for format SHORT_DATE for de_DE, de_LU, en, en_GB, fr_BE, fr_FR, fr_LU, nl_BE, nl_NL

chowlouise:
using yyyy for format SHORT_DATE for de_DE, de_LU, en, en_GB, fr_BE, fr_FR, fr_LU, nl_BE, nl_NL


fabalbon:
Closure's locale data is generated from CLDR (the industry standard locale data repository - http://cldr.unicode.org/). We have to keep Closure consistent with CLDR therefore to use yyyy for SHORT_DATE you should submit a request directly to CLDR (http://cldr.unicode.org/index/bug-reports) and it will be automatically updated into Closure when CLDR is updated.

Regards!

PR #621, opened 2015-12-18
Remove two dead pages from demos

mbeijen:
The Imageless Rounded Corder and Offline UI demo pages are included
in the demo page index but are not available - they lead to 404
errors. They describe functionality that has been removed from Closure
long ago.

The offline page was removed when goog.gears was removed from Closure in
7e9ce0ac84cc013055f5575b85b3147e046bbff2

The RoundedCorder demo page was removed in
1ab0fa8dff33aa92c384838edc918f47b68bbe6e
and in both cases the index was not modified.


mbeijen:
I signed it!

joeltine:
Thank you!

PR #625, opened 2015-12-22
Add ability for files to be ignored for the clang-format check.

joeltine:


joeltine:
FYI if anyone is curious, the "clang-format off" comments don't work for the clang-format-diff.py script as it doesn't parse the entire file.

scripts/ci/clang_ignore.txt
```diff
@@ -0,0 +1,9 @@
+closure/goog/locale/defaultlocalenameconstants.js
```
concavelenz: don't these need to turn off clang format internally too?  Shouldn't this be in the source?

joeltine: I could've *swore* clang-format-diff.py wasn't respecting the "// clang-format on|off" comments. This was supposed to be a quick workaround, but I just tested it now and it seems to be working. I will remove all of this clang_ignore stuff.

PR #628, opened 2016-01-04
Turn off clang-format check until Closure is fully formatted internally.

joeltine:


joeltine:
... and until the internal clang format check has had time to bake (just submitted today)

PR #640, opened 2016-01-28
Turkish currency is TRY not YTL

abides:
It fixes the Turkish currency name, ref: https://en.wikipedia.org/wiki/Turkish_lira (ISO 4217 code TRY (Numeric 949))


joeltine:
See https://github.com/google/closure-library/wiki/Internationalization-(i18n)-changes-in-Closure-Library

PR #641, opened 2016-01-28
Turkish shortdate format change

abides:
It should be dot seperator on short date not the space. ref: https://en.wikipedia.org/wiki/Date_and_time_notation_in_Turkey


joeltine:
See https://github.com/google/closure-library/wiki/Internationalization-(i18n)-changes-in-Closure-Library

PR #642, opened 2016-02-01
Fix node_modules glob pattern

joeltine:
Fix to adjust for regression mentioned in https://github.com/google/closure-compiler/issues/1407


Dominator008:
https://github.com/google/closure-compiler/pull/1480 Should fix that regression. Thanks for finding this out!

joeltine:
@Dominator008 Thanks for the quick turnaround.

PR #643, opened 2016-02-02
Disable the analyzer checks

MatrixFrog:
These are (much like the lint checks) not intended to be enabled as errors.


MatrixFrog:
This is my mistake by the way: When I split lintChecks and analyzerChecks into separate categories, I should have recommended this in the public release notes.

PR #644, opened 2016-02-02
calcdeps: process new goog.module() and goog.require() syntax

hochhaus:
Update the regular expressions in calcdeps.py to support the new goog.module() and goog.require() syntax. Even though calcdeps.py is deprecated it is still used for generation of jsdeps files by users of closure compiler modules. Currently, neither closurebuilder.py nor the compiler's manifest file can meet this use case:

https://github.com/google/closure-compiler/issues/1474

As this change is small would you consider a one time update to the legacy tool?

### DIFF
--- a/closure/bin/calcdeps.py
+++ b/closure/bin/calcdeps.py
@@ -43,9 +43,8 @@ import subprocess
 import sys
 
 
-_BASE_REGEX_STRING = '^\s*goog\.%s\(\s*[\'"](.+)[\'"]\s*\)'
-req_regex = re.compile(_BASE_REGEX_STRING % 'require')
-prov_regex = re.compile(_BASE_REGEX_STRING % 'provide')
+req_regex = re.compile('^(?:[^=]+=)?\s*goog\.require\(\s*[\'"](.+)[\'"]\s*\)')
+prov_regex = re.compile('^\s*goog\.(?:provide|module)\(\s*[\'"](.+)[\'"]\s*\)')
 ns_regex = re.compile('^ns:((\w+\.)*(\w+))$')
 version_regex = re.compile('[\.0-9]+')
 


closure/bin/calcdeps.py
```diff
@@ -43,9 +43,8 @@
 import sys
 
 
-_BASE_REGEX_STRING = '^\s*goog\.%s\(\s*[\'"](.+)[\'"]\s*\)'
-req_regex = re.compile(_BASE_REGEX_STRING % 'require')
-prov_regex = re.compile(_BASE_REGEX_STRING % 'provide')
+req_regex = re.compile('^(?:[^=]+=)?\s*goog\.require\(\s*[\'"](.+)[\'"]\s*\)')
+prov_regex = re.compile('^\s*goog\.(?:provide|module)\(\s*[\'"](.+)[\'"]\s*\)')
```
joeltine: I'm concerned about lack of external unit test on this change. It appears we have one internally, however. Because you signed the CLA, I can try pulling this internally and testing it.

hochhaus:
@joeltine If the internal test does not add enough coverage let me know and I'll add a test to this change. Thanks for having a look!

joeltine:
The test using goog.module was easy enough to add and it seems to work fine. I'll see about getting this out in the next couple days.

hochhaus:
Thanks @joeltine

joeltine:
Does [depswriter.py](https://github.com/google/closure-library/blob/master/closure/bin/build/depswriter.py) not do what you want? It's intended as the replacement for this script's deps file generation.

hochhaus:
Per my understanding depswriter.py does not have a mode equivalent to calcdeps.py's --output_mode=list. This mode is especially helpful for closure compiler --module flag file counting. The deps.js file output written by depswriter.py could require extra post processing to determine the order in which the files should be compiled.

joeltine:
Ah. Sorry. I didn't have full context on this change. 

I'm going to run this by the compiler folks first and make sure there's no existing workaround.

hochhaus:
Thanks for your help. From the discussion in https://github.com/google/closure-compiler/issues/1474 I don't think a  work around currently exists. I would be very happy to be proven wrong though.

joeltine:
MMkay, I think the main issue here is developing on the deprecated calcdeps.py. If list output mode is a missing feature, we should be adding it to depswriter.py instead of continuing to drag calcdeps along. Is that something you'd be willing to take on?

hochhaus:
That is a good point, I understand not wanting to update deprecated utilities. Regrettably, at the moment I don't have the time to look into adding a mode to depswriter.py. I honestly think that the compiler itself may be a better place than depswriter.py for long term solution.

When I have more time I'll loop back and if no support exists in the compiler I'll look into adding the mode to depswriter.

Thanks for your help.

hochhaus:
So I looked again and realized that closurebuilder.py does in fact have a --output_mode=list option (similar to the output in calcdeps.py). The problem is that, similar to the compiler manifest, the ordering of output dependencies is not well suited to module calculation. Therefore, when I get time to come back to this ticket, the action item is to either (a) fix the output ordering of the compilers manifest or (b) fix the output ordering of closurebuilders --output_mode=list.

PR #651, opened 2016-02-10
Remove a leftover internal URL from CONTRIBUTING

Dominator008:


shicks:
We'll merge a fix internally.  Thanks for pointing this out!

PR #655, opened 2016-02-12
Add "string" to return types of goog.array methods

Dominator008:
Namely, goog.find and goog.array.findRight can also return strings.

Found by Closure Compiler's NTI.

### DIFF
--- a/closure/goog/array/array.js
+++ b/closure/goog/array/array.js
@@ -496,8 +496,8 @@ goog.array.count = function(arr, f, opt_obj) {
  *     for every element. This function takes 3 arguments (the element, the
  *     index and the array) and should return a boolean.
  * @param {S=} opt_obj An optional "this" context for the function.
- * @return {T|null} The first array element that passes the test, or null if no
- *     element is found.
+ * @return {T|string|null} The first array element that passes the test, or
+ *     null if no element is found.
  * @template T,S
  */
 goog.array.find = function(arr, f, opt_obj) {
@@ -542,8 +542,8 @@ goog.array.findIndex = function(arr, f, opt_obj) {
  *     takes 3 arguments (the element, the index and the array) and should
  *     return a boolean.
  * @param {S=} opt_obj An optional "this" context for the function.
- * @return {T|null} The last array element that passes the test, or null if no
- *     element is found.
+ * @return {T|string|null} The last array element that passes the test, or
+ *     null if no element is found.
  * @template T,S
  */
 goog.array.findRight = function(arr, f, opt_obj) {


closure/goog/array/array.js
```diff
@@ -496,8 +496,8 @@ goog.array.count = function(arr, f, opt_obj) {
  *     for every element. This function takes 3 arguments (the element, the
  *     index and the array) and should return a boolean.
  * @param {S=} opt_obj An optional "this" context for the function.
- * @return {T|null} The first array element that passes the test, or null if no
- *     element is found.
+ * @return {T|string|null} The first array element that passes the test, or
```
shicks: Presumably the function is a function(string, number) which should set T to string already, no?

Also, it seems like making String implement IArrayLike<string> would maybe be a better way to go about fixing these issues.

Dominator008: Making string implement IArrayLike seems better. Would it be hard to land such a fix internally though?  @concavelenz

concavelenz:
I'd be tempted to just remove string as an in type

closure/goog/array/array.js
concavelenz: I'm really not excited to have string implement IArrayLike.   It seems more
likely to be an error than correct code.

Dominator008: `String` can implement `IArrayLike`, but not `string`.

concavelenz: I would rather prohibit string as an input than force string to be included
in the return type.

Dominator008: That does make it more in line with ES6's `Array.prototype.find`, if the internal cleanup isn't too bad.

concavelenz: Which?

Dominator008: https://github.com/google/closure-compiler/blob/master/externs/es6.js#L1424

I don't think "foo".find works.

concavelenz: Array.prototype.find.call("foo",...) would be the equivalent

Dominator008: Yeah I realized that...

concavelenz: I've made the change locally and will do a run to see if there is anything to be concerned about.

Dominator008: OK thanks.

Dominator008: @concavelenz Is there an update on this?

shicks:
Seems like this is a prime candidate for allowing "overloads" of type signatures.

Dominator008:
@shicks Yes, overloads would be ideal, but we need intersection types for that.

shicks:
Given that these methods are all but obsolete, and NTI is no longer around, I'm closing this PR.  Reopen if you feel we should still consider it.

PR #660, opened 2016-02-23
Allocate a new array for comparison in goog.array.stableSort

Dominator008:
This fixes a couple of warnings found by Closure Compiler's NTI. It should also be more performant on V8 because it avoids modifying the memory layout of "arr"'s backing store.

### DIFF
--- a/closure/goog/array/array.js
+++ b/closure/goog/array/array.js
@@ -1132,16 +1132,17 @@ goog.array.sort = function(arr, opt_compareFn) {
  * @template T
  */
 goog.array.stableSort = function(arr, opt_compareFn) {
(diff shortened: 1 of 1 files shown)


joeltine:
I couldn't find any discernible performance difference: http://jsperf.com/stablesort-perf/3.

OOC, what were the NTI errors?

Dominator008:
```
../closure-library/closure/goog/array/array.js:1136: WARNING - The right side in the assignment is not a subtype of the left side.
left side  : T
right side : {index:number, value:T}

    arr[i] = {index: i, value: arr[i]};
    ^
```

PR #662, opened 2016-02-23
Disable crossdomainrpc_test

joeltine:
Flaky on IE

### DIFF
--- a/protractor_spec.js
+++ b/protractor_spec.js
@@ -18,7 +18,8 @@ var IGNORED_TESTS = [
   'closure/goog/net/jsloader_test.html',
   'closure/goog/net/filedownloader_test.html',
   'closure/goog/promise/promise_test.html',
-  'closure/goog/editor/plugins/abstractdialogplugin_test.html'
+  'closure/goog/editor/plugins/abstractdialogplugin_test.html',
+  'closure/goog/net/crossdomainrpc_test.html'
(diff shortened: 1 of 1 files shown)


protractor_spec.js
```diff
@@ -18,7 +18,8 @@ var IGNORED_TESTS = [
   'closure/goog/net/jsloader_test.html',
   'closure/goog/net/filedownloader_test.html',
   'closure/goog/promise/promise_test.html',
-  'closure/goog/editor/plugins/abstractdialogplugin_test.html'
+  'closure/goog/editor/plugins/abstractdialogplugin_test.html',
```
concavelenz: I'm a little unclear on why things are continually being added to this list.

joeltine: They're the tests that have promise timeouts that fail before the set timeout value. I just add them to this list as they come up.

joeltine: Obviously, this is meant to be temporary.

Dominator008: I wonder if `closure/goog/editor/plugins/abstractdialogplugin_test.html` is fixed by the following commit:
https://github.com/google/closure-library/commit/a1b20191b1cfd7937b1a3402eb1347bd54bc3116#diff-c0a7f4c78a87dc0d9c753419bf2fd040R89

PR #664, opened 2016-02-26
Add Customized Splitter instead of hard code character '-'

SimonCat1989:
Add "@define goog.CSS_CLASS_SPLITTER" in base.js from closure library to make function goog.getCssName in base.js can accept customized css name splitter instead of original hard code character "-".


concavelenz:
What is the use case for this?  Were you intending to update the other tools as well?

joeltine:
Closing due to lack of response from author.

PR #669, opened 2016-03-03
Fix insertAll logic when heap is empty

Kolo9:
Fixed logic for insertAll into empty heap, added insertAll test cases, fixed parameter order of assertEquals calls in heap tests

### DIFF
--- a/closure/goog/structs/heap_test.js
+++ b/closure/goog/structs/heap_test.js
@@ -19,67 +19,67 @@ goog.require('goog.structs');
 goog.require('goog.structs.Heap');
 goog.require('goog.testing.jsunit');
 
-function getHeap() {
-  var h = new goog.structs.Heap();
-  h.insert(0, 'a');
-  h.insert(1, 'b');
-  h.insert(2, 'c');
-  h.insert(3, 'd');
-  return h;
-}
 
-
-function getHeap2() {
+/**
+ * Constructs a heap from key-value pairs passed as arguments
+ * @param {...!Array} List of length 2 arrays [key, value] as arguments
+ * @returns {goog.structs.Heap} Heap constructed from passed in key-value pairs
+ */
+function makeHeap() {
   var h = new goog.structs.Heap();
-  h.insert(1, 'b');
-  h.insert(3, 'd');
-  h.insert(0, 'a');
-  h.insert(2, 'c');
+  var key, value;
+
+  for (var i = 0; i < arguments.length; i++) {
+    key = arguments[i][0];
+    value = arguments[i][1];
+    h.insert(key, value);
+  }
+
   return h;
 }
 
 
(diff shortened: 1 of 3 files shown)


closure/goog/structs/heap_test.js
```diff
@@ -205,11 +205,39 @@ function testInsertPeek2() {
...
-  assertEquals('peak, Should be "a"', h.peek(), 'a');
+  assertEquals('peek, Should be "a"', 'a', h.peek());
   h.insert(2, 'c');
-  assertEquals('peak, Should be "a"', h.peek(), 'a');
+  assertEquals('peek, Should be "a"', 'a', h.peek());
+}
+
+function testInsertAllPeek1() {
+  var h1 = new goog.structs.Heap();
+  var h2 = getHeap();
```
shicks: This test is difficult to read and understand because test logic that is important to the correctness of its assertions is external to the actual method.  This is not your fault - getHeap was here when you started - but I would be a lot happier about it if we had, say, a makeHeap method that took pairs and inserted them in the right order:

  var h1 = makeHeap();
  var h2 = makeHeap([0, 'a'], [2, 'c'], [1, 'b']);

This would make it completely obvious from just the method (and the "obvious" implementation of makeHeap) that the test was correct.  Would you mind pulling this method out as well?

Kolo9: Absolutely! I wasn't fond of it, either, but figured I'd stay consistent with the rest of the cases. I'll change the tests up as you mentioned to make them more readable.

closure/goog/structs/heap_test.js
```diff
@@ -130,86 +126,114 @@ function testIsEmpty() {
...
   h.insert(2, 'c');
-  assertEquals('peak, Should be "a"', h.peek(), 'a');
+  assertEquals('peek, Should be "a"', 'a', h.peek());
+}
+
+function testInsertAllPeek1() {
+  var h1 = makeHeap();
+  var h2 = makeHeap([0, 'a'], [1, 'b'], [2, 'c'], [3, 'd']);
+
+  h1.insert(1, 'e');
```
shicks: h1 = makeHeap(1, 'e');

closure/goog/structs/heap_test.js
```diff
@@ -19,67 +19,63 @@ goog.require('goog.structs');
 goog.require('goog.structs.Heap');
 goog.require('goog.testing.jsunit');
 
-function getHeap() {
+
+//Takes list of length 2 arrays [key, value] as arguments
```
shicks: Our style guide requires a space after the //

shicks:
Thanks for the contribution - in order for us to merge it, could you please squash these commits and rebase it on top of our master branch and then `git push -f` to rewrite your history (rather than merging).

PR #671, opened 2016-03-04
Fix a few lint warnings.

Dominator008:


closure/goog/testing/performancetimer.js
```diff
@@ -194,6 +194,7 @@ goog.testing.PerformanceTimer.prototype.runTask = function(task) {
  * Finishes the run of a task by creating a result object from samples, in the
  * format described in {@code run}.
  * @see goog.testing.PerformanceTimer#run
+ * @param {Array<number>} samples The samples to analyze.
```
shicks: We should add a lint warning about implicit nullability...

Dominator008: @shicks That's part of analyzerChecks which is off for Closure Library now: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/DiagnosticGroups.java#L507.

closure/goog/testing/propertyreplacer.js
```diff
@@ -116,6 +116,7 @@ goog.testing.PropertyReplacer.hasKey_ = function(obj, key) {
     //   var dummy = document.body.constructor.prototype.className
     //   [Exception... "Cannot modify properties of a WrappedNative"]
     // </pre>
+    /** @suppress {unusedLocalVariables} */
```
shicks: Internal linter currently complains about this line, saying "@suppress is not allowed here".  Is that changing?  Can we land the linter change first?

Dominator008: I removed that suppression for now. In fact most of the lint warnings in Closure Library are "@suppress is not allowed here". I think we need to allow more suppressions in the Compiler: https://github.com/google/closure-compiler/issues/1615.

shicks: Alright, sounds good - we can fix those up once they're allowed.

PR #672, opened 2016-03-07
Fix a few more lint warnings.

Dominator008:
Fix goog.provide()'s that were not sorted alphabetically in closure/goog/proto2/test.pb.js.

Add missing JSDoc to functions in closure/goog/soy/soy_testhelper.js.

### DIFF
--- a/closure/goog/soy/soy_testhelper.js
+++ b/closure/goog/soy/soy_testhelper.js
@@ -51,8 +51,16 @@ function SanitizedContentSubclass(content, kind) {
 goog.inherits(SanitizedContentSubclass, goog.soy.data.SanitizedContent);
 
 
+/**
+ * @param {string} content The text.
+ * @param {goog.soy.data.SanitizedContentKind|string} kind The kind of safe
+ *     content.
+ * @return {!SanitizedContentSubclass}
+ */
 function makeSanitizedContent(content, kind) {
-  return new SanitizedContentSubclass(content, kind);
+  return new SanitizedContentSubclass(
+      content,
+      /** @type {goog.soy.data.SanitizedContentKind} */ (kind));
 }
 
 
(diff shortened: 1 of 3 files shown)


Dominator008:
This might require an internal cleanup if some teams are still on gjslint.

concavelenz:
I think it is worth having a conversation about whether this is something the compile based linter is going to add first.

Dominator008:
@MatrixFrog

MatrixFrog:
We're already having that conversation in https://github.com/google/closure-compiler/issues/1615

Dominator008:
@MatrixFrog I think what @concavelenz meant is that whether the Compiler-based linter should actually check for "extraProvide", not just whether it should be allowed (and essentially ignored).

Dominator008:
@concavelenz I'm adding `extraProvide` to allowed suppression types in the Compiler-based linter in https://github.com/google/closure-compiler/pull/1624. I've reduced this PR to just fixing some lint warnings.

closure/goog/soy/soy_testhelper.js
```diff
@@ -64,44 +72,79 @@ function makeSanitizedContent(content, kind) {
 var example = {};
 
 
-example.textNodeTemplate = function(opt_data, opt_sb, opt_injectedData) {
-  assertNotNull(opt_data);
-  assertNotUndefined(opt_data);
-  return goog.string.htmlEscape(opt_data.name);
+/**
+ * @param {{name: string}} data
+ * @param {null=} opt_sb
```
joeltine: Why `{null=}` here? This type wizardy scares and confuses me.

Dominator008: I think it is a legacy [parameter or something.. I can't figure out its full name.

joeltine: How did you determine these `@param` types?

Dominator008: These are Soy templates.
`soy_test.js` uses:
`goog.soy.renderElement`: https://github.com/google/closure-library/blob/master/closure/goog/soy/soy.js#L67-L68
`goog.soy.renderAsFragment`: https://github.com/google/closure-library/blob/master/closure/goog/soy/soy.js#L91-L92
I determined the types from the signatures there.

PR #677, opened 2016-03-15
Added goog.structs.Map.prototype.findValue and findKey

ghosttie:
Added goog.structs.Map.prototype.findValue and goog.structs.Map.prototype.findKey to add functionality missing when converting an Object<K, V> to a Map<K, V>.

These are simply modified versions of goog.object.findValue and goog.object.findKey.

See discussion [here](https://groups.google.com/forum/#!topic/closure-compiler-discuss/vGdVqmcr8n4)

cc @concavelenz

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -23,3 +23,4 @@ John Huân Vũ <jvu.calpoly@gmail.com>
 Peter Lu <peterlu83+github@gmail.com>
 Filipe Catraia <filipe.catraia@deliveryhero.com>
 Dan Rubalsky <drubalsky@plentakill.com>
+FileVision USA LLC <jmunro@filevision.net>
--- a/closure/goog/structs/map.js
+++ b/closure/goog/structs/map.js
@@ -350,6 +350,47 @@ goog.structs.Map.prototype.forEach = function(f, opt_obj) {
 
 
 /**
+ * Searches the map for an element that satisfies the given condition and
+ * returns its key.
+ * @param {function(this:T,V,string,goog.structs.Map<K,V>):boolean} f The
+ *      function to call for every element. Takes 3 arguments (the value,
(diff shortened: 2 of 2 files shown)


ghosttie:
I don't see how my change could have made testExistingImage timeout

Dominator008:
@ghosttie I think that's a known failure not related to your change.

joeltine:
@ghosttie I think these API additions would be better off as a general structs utility. That is, a method in https://github.com/google/closure-library/blob/master/closure/goog/structs/structs.js. This may be more than you'd like to take on though.

ghosttie:
OK I only did it that way because Map seemed to use member functions rather than utility functions e.g. goog.array.forEach vs goog.structs.Map.prototype.forEach

joeltine:
@MatrixFrog Do you recall why goog.structs.forEach was slated for deprecation and Map has its own implementation?

MatrixFrog:
goog.structs.forEach is just a bit weird. It's handling maps, sets, arrays, and any collection that happens to have a method called 'forEach', all in one function. It's rare that you want to iterate over something but have no idea what the thing is. If you know it's a map, use the map's own forEach method. Likewise with goog.structs.get{Keys,Values}.

shicks:
I'm not sure we actually want to be encouraging the use of `goog.structs.Map` at all.  Now that the compiler supports ES6 `Map`, that's really a better option in basically every situation.  `goog.structs.Map` has a number of fundamental problems (primarily around non-string keys) and gives users a false sense of confidence that it will "do the right thing" when it really won't.  I'm hopeful that it will be officially deprecated Any Day Now™.

joeltine:
Mainly for the reasons expressed in the previous comment, I'm closing this out.

PR #574, opened 2015-10-01
Refine type returns of createElement()

psam44:
For https://github.com/google/closure-library/issues/570

PR #678, opened 2016-03-16
fix for WindowsError: [Error 206] The filename or extension is too long

ikabirov:


ikabirov:
I signed it!

ikabirov:
I signed it!

closure/bin/build/jscompiler.py
```diff
@@ -97,12 +97,17 @@ def _GetJsCompilerArgs(compiler_jar_path, java_version, source_paths,
...
 
-  for path in source_paths:
-    args += ['--js', path]
+  #collect list of compiler flags
+  compiler_flags.extend(['--js ' + path for path in source_paths])
+  if jvm_flags:
+    compiler_flags.extend(jvm_flags)
+
+  #write flags to file for avoiding problem with too big arguments length
+  flags_file = tempfile.NamedTemporaryFile(delete = False)
```
joeltine: `delete = False` means this file won't be deleted even after the program exits, correct? I'd prefer not to pollute the user's disk as someone could be compiling very often.

Can you write this in such a way that it cleans up after itself? For example, rm'ing the file after compilation.

joeltine:
Where exactly is `WindowsError: [Error 206]` coming from? From the error description, it only has to do with filenames. How does args equate to filenames or extensions?

closure/bin/build/jscompiler.py
ikabirov: Fixed

ikabirov:
`Traceback (most recent call last):
  File "C:\projects\libs\closure\closure\bin\build\closurebuilder.py", line 287, in <module>
    main()
  File "C:\projects\libs\closure\closure\bin\build\closurebuilder.py", line 276, in main
    compiler_flags=options.compiler_flags)
  File "...\libs\closure\closure\bin\build\jscompiler.py", line 133, in Compile
    return subprocess.check_output(args)
  File "C:\Python27\lib\subprocess.py", line 566, in check_output
    process = Popen(stdout=PIPE, *popenargs, **kwargs)
  File "C:\Python27\lib\subprocess.py", line 710, in __init__
    errread, errwrite)
  File "C:\Python27\lib\subprocess.py", line 958, in _execute_child
    startupinfo)
WindowsError: [Error 206] The filename or extension is too long`
Python fails to create process on big projects. Temp file on our projects can reach 100-150 KB

closure/bin/build/jscompiler.py
```diff
@@ -97,12 +97,16 @@ def _GetJsCompilerArgs(compiler_jar_path, java_version, source_paths,
   # Add the application JAR.
   args += ['-jar', compiler_jar_path]
 
-  for path in source_paths:
-    args += ['--js', path]
+  #collect list of compiler flags
+  compiler_flags.extend(['--js ' + path for path in source_paths])
```
joeltine: Please stay consistent with the `+=` (compared to `extend()`) usage already in the file

closure/bin/build/jscompiler.py
```diff
@@ -97,12 +97,16 @@ def _GetJsCompilerArgs(compiler_jar_path, java_version, source_paths,
   # Add the application JAR.
   args += ['-jar', compiler_jar_path]
 
-  for path in source_paths:
-    args += ['--js', path]
+  #collect list of compiler flags
```
joeltine: Space after `#`. Also, properly capitalize and end comments with a period. Here and throughout.

closure/bin/build/jscompiler.py
```diff
@@ -69,7 +69,7 @@ def _JavaSupports32BitMode():
   return supported
 
 
-def _GetJsCompilerArgs(compiler_jar_path, java_version, source_paths,
+def _GetJsCompilerArgs(compiler_jar_path, java_version, source_paths, flags_file,
```
joeltine: We have an 80 char line length limit. I can't tell if this line or :133 are over 80.

closure/bin/build/jscompiler.py
joeltine: `compiler_flags` can be `None` here and throw. You'll need to initialize it in that case.

closure/bin/build/jscompiler.py
```diff
@@ -97,12 +97,16 @@ def _GetJsCompilerArgs(compiler_jar_path, java_version, source_paths,
   # Add the application JAR.
   args += ['-jar', compiler_jar_path]
 
-  for path in source_paths:
-    args += ['--js', path]
+  #collect list of compiler flags
+  compiler_flags.extend(['--js ' + path for path in source_paths])
+  if jvm_flags:
+    compiler_flags.extend(jvm_flags)
```
joeltine: I'm confused. Why are you re-adding `jvm_flags` to the arg list? This is redundant as they're added above.

closure/bin/build/jscompiler.py
```diff
@@ -124,12 +128,16 @@ def Compile(compiler_jar_path, source_paths,
 
   java_version = _ParseJavaVersion(str(_GetJavaVersionString()))
 
+  flags_file = tempfile.NamedTemporaryFile(delete = False)
   args = _GetJsCompilerArgs(
-      compiler_jar_path, java_version, source_paths, jvm_flags, compiler_flags)
+      compiler_jar_path, java_version, source_paths, flags_file, jvm_flags, compiler_flags)
```
joeltine: Passing a `flags_file` handle feels very strange here. I think a better pattern would be something like:

```
args = getjscompilerargs(...)
flag_file = getflagfile(....)
args += ['--flag_file', flag_file.name]
```

With this you have a nice separation of concerns and you're not left guessing what `GetJsCompilerArgs` does with a flags file you pass it.

closure/bin/build/jscompiler.py
```diff
@@ -124,12 +128,16 @@ def Compile(compiler_jar_path, source_paths,
...
   logging.info('Compiling with the following command: %s', ' '.join(args))
 
   try:
-    return subprocess.check_output(args)
+    result = subprocess.check_output(args)
+    os.remove(flags_file.name)
+    return result
   except subprocess.CalledProcessError:
-    raise JsCompilerError('JavaScript compilation failed.')
+    os.remove(flags_file.name)
```
joeltine: You can just remove the file in a `finally` block

closure/bin/build/jscompiler.py
joeltine: ... also please add a comment why we're using `--flag_file`.

closure/bin/build/jscompiler.py
```diff
@@ -97,15 +96,22 @@ def _GetJsCompilerArgs(compiler_jar_path, java_version, source_paths,
...
   args += ['-jar', compiler_jar_path]
 
+  return args
+
+def _GetFlagFile(source_paths, compiler_flags):
+  args = []
   for path in source_paths:
     args += ['--js', path]
-
+
```
joeltine: Rm trailing whitespace

closure/bin/build/jscompiler.py
```diff
@@ -125,7 +131,12 @@ def Compile(compiler_jar_path, source_paths,
...
   args = _GetJsCompilerArgs(
-      compiler_jar_path, java_version, source_paths, jvm_flags, compiler_flags)
+      compiler_jar_path, java_version, jvm_flags)
+
+  # Write source path arguments to flag file for avoiding "The filename or 
+  # extension is too long" error in big projects.
+  flags_file = _GetFlagFile(source_paths, compiler_flags)
+  args += ['--flagfile', flags_file.name]
 
   logging.info('Compiling with the following command: %s', ' '.join(args))
```
joeltine: It's unfortunate we lose the logging of all the the `--js` flags here. I honestly don't know how much people rely on this output. We can just try it as-is for now and see if anyone complains.

closure/bin/build/jscompiler.py
```diff
@@ -125,7 +131,12 @@ def Compile(compiler_jar_path, source_paths,
   java_version = _ParseJavaVersion(str(_GetJavaVersionString()))
 
   args = _GetJsCompilerArgs(
-      compiler_jar_path, java_version, source_paths, jvm_flags, compiler_flags)
+      compiler_jar_path, java_version, jvm_flags)
+
+  # Write source path arguments to flag file for avoiding "The filename or 
+  # extension is too long" error in big projects.
```
joeltine: Add a link to this PR as a reference please.

joeltine:
Also, you'll need to squash your commits to one. See https://github.com/google/closure-library/pull/574#issuecomment-146009727 and http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

closure/bin/build/jscompiler.py
```diff
@@ -69,8 +69,7 @@ def _JavaSupports32BitMode():
   return supported
 
 
-def _GetJsCompilerArgs(compiler_jar_path, java_version, source_paths,
-                       jvm_flags, compiler_flags):
+def _GetJsCompilerArgs(compiler_jar_path, java_version, jvm_flags):
```
joeltine: You'll need to update the unit test [jscompiler_test.py](https://github.com/google/closure-library/blob/master/closure/bin/build/jscompiler_test.py) to reflect these changes.

ikabirov:
What happened with Travis CI build?

joeltine:
This error is from a Closure Compiler bug. They're working on it.

```
./closure/goog/test_module.js:25: ERROR - required "goog.test_module_dep" namespace never provided
goog.require('goog.test_module_dep');
```

The other errors were from Sauce Labs. It reached its VM limit (due to multiple runs). A test rerun usually solves this.

PR #683, opened 2016-03-22
Write bytes to stdout when using Python 3

keemyb:
Fixes "TypeError: can't concat bytes to str"

### DIFF
--- a/closure/bin/build/closurebuilder.py
+++ b/closure/bin/build/closurebuilder.py
@@ -34,6 +34,7 @@ import logging
 import optparse
 import os
 import sys
+import io
 
 import depstree
 import jscompiler
@@ -200,9 +201,14 @@ def main():
 
   # Make our output pipe.
   if options.output_file:
-    out = open(options.output_file, 'w')
+    out = io.open(options.output_file, 'wb')
   else:
-    out = sys.stdout
+    version = sys.version_info[:2]
+    if version >= (3,0):
+        # Write bytes to stdout
+        out = sys.stdout.buffer
+    else:
+        out = sys.stdout
 
   sources = set()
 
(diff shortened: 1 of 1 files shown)


closure/bin/build/closurebuilder.py
```diff
@@ -276,7 +282,7 @@ def main():
         compiler_flags=options.compiler_flags)
 
     logging.info('JavaScript compilation succeeded.')
-    out.write(str(str(compiled_source).encode('utf-8')))
+    out.write((compiled_source).encode('utf-8'))
```
joeltine: do you need the parens around `compiled_source`?

closure/bin/build/closurebuilder.py
```diff
@@ -200,9 +201,14 @@ def main():
 
   # Make our output pipe.
   if options.output_file:
-    out = open(options.output_file, 'w')
+    out = io.open(options.output_file, 'wb')
   else:
-    out = sys.stdout
+    version = sys.version_info[:2]
+    if version >= (3,0):
```
joeltine: At a glance, this change seems fine to me. There's a bigger underlying question here of python 3 support in general for our tools. This may fix this one issue, but I'm sure there are other places in the code that aren't Py3 compliant. Also, I don't have a lot of experience making scripts both Py3 and Py2 compatible, but I'd hate to see this type of pattern repeated all over the place (i.e., `if py3: do blah else: do foo`).

Let me chat with some other folks about this.

closure/bin/build/closurebuilder.py
keemyb: No, that looks like a mistake on my part.

closure/bin/build/closurebuilder.py
joeltine: OK, I think we're OK merging these types of fixes.

closure/bin/build/closurebuilder.py
joeltine: Can you remove them?

joeltine:
Can you please squash your commits? http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

PR #684, opened 2016-03-23
Handle touch events in goog.fx.AbstractDragDrop

codyebberson:
The current implementation of AbstractDragDrop explicitly listens
for mousedown, mousemove, and mouseup events.  Therefore it does
not work with touch devices.

This limitation can easily be observed in the dragdrop.html demo.

This commit adds touchstart, touchmove, and touchend support to the
AbstractDragDrop class.

### DIFF
--- a/closure/goog/fx/abstractdragdrop.js
+++ b/closure/goog/fx/abstractdragdrop.js
@@ -295,8 +295,9 @@ goog.fx.AbstractDragDrop.prototype.init = function() {
 goog.fx.AbstractDragDrop.prototype.initItem = function(item) {
   if (this.isSource_) {
     goog.events.listen(
-        item.element, goog.events.EventType.MOUSEDOWN, item.mouseDown_, false,
-        item);
+        item.element,
+        [goog.events.EventType.MOUSEDOWN, goog.events.EventType.TOUCHSTART],
+        item.dragStartHandler_, false, item);
     if (this.sourceClass_) {
       goog.dom.classlist.add(
           goog.asserts.assert(item.element), this.sourceClass_);
@@ -319,8 +320,9 @@ goog.fx.AbstractDragDrop.prototype.initItem = function(item) {
 goog.fx.AbstractDragDrop.prototype.disposeItem = function(item) {
   if (this.isSource_) {
     goog.events.unlisten(
-        item.element, goog.events.EventType.MOUSEDOWN, item.mouseDown_, false,
-        item);
+        item.element,
+        [goog.events.EventType.MOUSEDOWN, goog.events.EventType.TOUCHSTART],
+        item.dragStartHandler_, false, item);
     if (this.sourceClass_) {
       goog.dom.classlist.remove(
           goog.asserts.assert(item.element), this.sourceClass_);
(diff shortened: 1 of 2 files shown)


shicks:
Looks reasonable.  I'm going to try to push this through our internal presubmit and see if anything breaks.

shicks:
Presubmit passed.  We'll get another set of eyes on it and then should be able to check it in.

closure/goog/fx/abstractdragdrop.js
```diff
@@ -319,8 +320,9 @@ goog.fx.AbstractDragDrop.prototype.initItem = function(item) {
 goog.fx.AbstractDragDrop.prototype.disposeItem = function(item) {
   if (this.isSource_) {
     goog.events.unlisten(
-        item.element, goog.events.EventType.MOUSEDOWN, item.mouseDown_, false,
-        item);
+        item.element,
+        [goog.events.EventType.MOUSEDOWN, goog.events.EventType.TOUCHSTART],
+        item.mouseDown_, false, item);
```
joeltine: I don't think this will work. You bind both MOUSEDOWN and TOUCHSTART to mouseDown_ listener, but mouseDown_ intentionally returns on TOUCHSTART. This will break basically everything as the "start drag" event will never fire. 

In general, this change needs to do some refactoring to abstract the concept of "mouse" out of  these methods.

codyebberson:
@shicks @joeltine - Thanks for looking at this.

> intentionally returns on TOUCHSTART

The opposite actually.  I modified that early exit to _allow_ TOUCHSTART.  The mouseDown_ method had an early exit on ```!event.isMouseActionButton()``` to check for "non-action" cases such as right click and ctrl+click.  I want to allow TOUCHSTART as another "action" case.

I considered modifying isMouseActionButton to allow touchstart, but (1) didn't want to increase the footprint of the pull request and (2) felt like that was not in the spirit of a function with "mouse" in the name.

> In general, this change needs to do some refactoring to abstract the concept of "mouse" out of these methods.

Perhaps a "isActionButton" method such as:
```
function isActionButton(e) {
    return e.isMouseActionButton() || e.type !== goog.events.EventType.TOUCHSTART;
}
```

Or did you have something more comprehensive in mind?

> I don't think this will work.

We have been using this patch in production for over a year, and it has behaved as expected for our use cases.  

Your comment did lead me to reconsider assumptions.  The one case that might get messy is a touchscreen laptop where the user can both click and touch.  You could get into a weird state under those circumstances.  I believe that is an acceptable edge case, and that this is still a positive improvement.

joeltine:
> The opposite actually.

Whoops, sorry about that. I misread that `mouseDown_` statement.

> Or did you have something more comprehensive in mind?

Something more comprehensive. Just lobbing touch events into a `mouseDown_` method feels forced. All of the comments mention mouse specifically. 

We just need better naming on those mouse specific methods. Maybe `dragStartHandler_` `dragStopHandler_`?

codyebberson:
> We just need better naming on those mouse specific methods. Maybe dragStartHandler_ dragStopHandler_?

Sure, understood.  Stay tuned.

codyebberson:
Quick bump that suggested changes have been submitted.

joeltine:
I just pulled in the new version internally. I'll run some tests and update the thread.

joeltine:
FYI: I've had to this change back internally. The change will make it out to the github repo at head either today or tomorrow.

The reasoning is:

"This interaction makes it so the user cant scroll since for drag targets that cover the whole scrollable area. Plus this is not the conventional way to drag and drop on touch. Conventionally, you need to long-press, and this one drags immediately"

This feature will probably need to be made configurable.

PR #709, opened 2016-05-26
Add optional context to the withResolver function.

pcj:
The `withResolver` method should take an optional context parameter.  This does not break any existing code.


pcj:
CLA signed.  Soul accepted by devil.

pcj:
Will Retry

PR #710, opened 2016-05-26
Adds the opt_context parameter to the withResolver function.

pcj:
The withResolver function should accept a context object for inclusion into the constructed promise object.

### DIFF
--- a/closure/goog/promise/promise.js
+++ b/closure/goog/promise/promise.js
@@ -515,17 +515,20 @@ goog.Promise.firstFulfilled = function(promises) {
 
 
 /**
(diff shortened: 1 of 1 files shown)


pcj:
Looks like build is failing on master despite this change.  Any suggestions?

joeltine:
Ignore it. It's a sauce labs bug.

shicks:
Thanks for your interest.

The context only matter if the first argument (`resolver`) passed into `new goog.Promise` refers to `this`.  The `resolver` in `withResolver` only refers to local variables, so the context doesn't actually play a role here.

pcj:
Hmm.  By gum you're right!  Thanks for the explanation.

PR #715, opened 2016-06-01
Move goog.loadModule{,FromSource_} out of DEPENDENCIES_ENABLED guard

### LINKED ISSUE #711: Support WHITESPACE_ONLY mode with goog.ENABLE_DEBUG_LOADER=false
As of the `v20160517` compiler release when compiling in `WHITESPACE_ONLY` mode the `WhitespaceWrapGoogModules` compile pass wraps `goog.module()` files in calls to `goog.loadModule(...)`. However, when setting `goog.ENABLE_DEBUG_LOADER=false` base.js does not define `goog.loadModule()`. This makes the combination of WHITESPACE_ONLY mode and `goog.ENABLE_DEBUG_LOADER=false` currently incompatible.

I originally filed https://github.com/google/closure-compiler/issues/1815 about this issue. @MatrixFrog suggested that a library level fix (moving the definition of `goog.loadModule()` out of the `goog.DEPENDENCIES_ENABLED` conditional and guarding it on another condition) might be the best course of action.

 @concavelenz @jart

hochhaus:
Fixed #711

### DIFF
--- a/closure/goog/base.js
+++ b/closure/goog/base.js
@@ -812,6 +812,68 @@ goog.define('goog.ALWAYS_TRANSPILE', false);
 goog.define('goog.NEVER_TRANSPILE', false);
 
 
+/**
+ * @param {function(?):?|string} moduleDef The module definition.
+ */
+goog.loadModule = function(moduleDef) {
+  // NOTE: we allow function definitions to be either in the from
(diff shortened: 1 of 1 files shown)


hochhaus:
Fixed by bee7888290515eaf239abff519da8841e55cd5c0.

PR #718, opened 2016-06-16
Hardcode Chrome version to maintain compatibility

joeltine:


joeltine:
Fixes issues like https://travis-ci.org/google/closure-library/builds/137966450 where we've fixed stuff for a newer version of Chrome but the default version is behind.

PR #596, opened 2015-11-04
Ensuring MatrixConstructor works for Firefox

Antonio-Laguna:
Currently the MatrixConstructor doesn't really work with Firefox, this fixes the issue.

PR #719, opened 2016-06-22
Remove use of CSSMatrix constructor.

shans:
CSSMatrix was a non-standardized constructor that was briefly slated for inclusion in CSS Transitions Level 3 (see https://developer.mozilla.org/en-US/docs/Web/API/CSSMatrix). No current browser ships CSSMatrix (Chrome & Safari ship WebKitCSSMatrix, IE & Edge ship MSCSSMatrix).

The CSS Houdini TypedOM specification (https://www.w3.org/TR/css-typed-om-1/) has introduced CSSMatrix as a CSSTransformComponent constructor (https://www.w3.org/TR/css-typed-om-1/#cssmatrix) and support is in the process of being added to the Blink renderer.

This means that, right now, if Chrome is executed with experimental web platform features switched on, sites that attempt to use the CSSMatrix constructor through closure do not work correctly.

This patch should resolve that issue - site can continue to use the existing WebKitCSSMatrix or MSCSSMatrix constructors, and the new CSSMatrix object won't be invoked.

### DIFF
--- a/closure/goog/style/transform.js
+++ b/closure/goog/style/transform.js
@@ -164,8 +164,5 @@ goog.style.transform.matrixConstructor_ =
       if (goog.isDef(goog.global['MSCSSMatrix'])) {
         return goog.global['MSCSSMatrix'];
       }
-      if (goog.isDef(goog.global['CSSMatrix'])) {
-        return goog.global['CSSMatrix'];
-      }
       return null;
     });


shicks:
I'm not clear on why this breaks newer browsers - they should bail out early with either `WebKitCSSMatrix` or `MSCSSMatrix`, and then this removed code is ever executed.  Am I missing something?

Also, we have another PR #596 that adds `DomMatrix` - where does that fit into this?

joeltine:
Closing due to lack of response.

PR #724, opened 2016-06-29
Added additional checks to avoid error messages for edge cases.

MarkusBordihn:


MarkusBordihn:
After several different combinations of `@suppress`, I have no idea why the following errors still showing up on Travis CI:
```
./closure/goog/graphics/vmlelement.js:120: ERROR - missing require: 'goog.graphics.VmlGraphics'

      goog.graphics.VmlGraphics.toSizeCoord(height);

      ^

./closure/goog/graphics/vmlelement.js:278: ERROR - missing require: 'goog.graphics.VmlGraphics'

      goog.graphics.VmlGraphics.toSizePx(height);

      ^

./closure/goog/graphics/vmlelement.js:426: ERROR - missing require: 'goog.graphics.VmlGraphics'

      goog.graphics.VmlGraphics.toPosPx(height);

      ^

3 error(s), 0 warning(s)
```

I'm unable to reproduce these errors locally.

closure/goog/graphics/vmlelement.js
```diff
@@ -109,11 +109,16 @@ goog.graphics.VmlGroupElement.prototype.setSize = function(width, height) {
   var element = this.getElement();
 
   var style = element.style;
-  style.width = goog.graphics.VmlGraphics.toSizePx(width);
-  style.height = goog.graphics.VmlGraphics.toSizePx(height);
-
-  element.coordsize = goog.graphics.VmlGraphics.toSizeCoord(width) + ' ' +
+  style.width = /** @suppress {missingRequire} */
```
shicks: @suppress {missingRequire} does not work in the middle of an expression.  At the very narrowest it should be on the statement, but having it on the function is also fine, particularly since it can be more clearly documented there.

closure/goog/ui/tree/basenode.js
```diff
@@ -186,16 +186,16 @@ goog.ui.tree.BaseNode.prototype.initAccessibility = function() {
     }
 
     var img = this.getIconElement();
-    if (img) {
+    if (img && img.setAttribute) {
```
shicks: What edge case is this intended to address?  Shouldn't the return type of getIconElement already account for it, or is there something missing?

.gitignore
```diff
@@ -11,4 +11,9 @@
 # python
 *.pyc
 
+# Node JS
 node_modules/
+npm-debug.log
+
+# General
+.DS_Store
```
shicks: Seems like these should be a separate PR.

closure/goog/ui/tree/basenode.js
MarkusBordihn: I had the case that getExpandIconElement(), getIcontElement(), ... returned no real element like an &nbsp;  (which has not attribute or style).

Issue #731, opened 2016-07-15
goog.ui.Container.prototype.setHighlightedIndex doesn't set this.highlightedIndex_


ghosttie:
goog.ui.Container.prototype.setHighlightedIndex doesn't set this.highlightedIndex_ so calling goog.ui.Container.prototype.getHighlightedIndex doesn't return the index that was just set

joeltine:
Any chance of a PR? goog.ui is mostly in maintenance mode atm.

ghosttie:
https://github.com/google/closure-library/pull/733

PR #733, opened 2016-07-18
Fix highlightedIndex_ not being set in setHighlightedIndex

ghosttie:
As discussed in https://github.com/google/closure-library/issues/731


ghosttie:
I'm not sure I understand why the The Travis CI build failed

mashedcode:
@ghosttie well it's because if `setHighlightedIndex(-1)` is called the highlight should be removed. Obviously no child at the index `-1` is found and `this.highlightedIndex_ > -1` will always evaluate to `false` in that case and because of your change. Therefore you can't remove the current highlight anymore and the test fails.

You obviously only want to set the highlighted index to something `> -1` if there's actually a child that will be highlighted anyway so that's another issue.

ghosttie:
@mashedcode thanks, that was very helpful - I'll be more self-sufficient next time

ghosttie:
I don't think 

> Failed: Modal dialog present: The Acrobat DC PDF browser plug-in is missing.

is my fault...

joeltine:
Sauce Labs *shakes fist*!

That failure is unrelated to this CL. It's a Sauce Labs infra platform issue.

ghosttie:
How about

> Failed: ERROR The test with session id c52b1bf454734499adc752c66df1edcd has already finished, and can't receive further commands.

joeltine:
Same. It means we ran out of VMs on Sauce Labs. It typically happens when multiple PRs/commits come in at the same time.

ghosttie:
Does it automatically retry or do I need to do something?

joeltine:
I restarted it.

ghosttie:
> closure/goog/net/streams/jsonstreamparser_test.html timed out after 45s!

This is less fun that I was expecting

joeltine:
I wouldn't be too concerned getting Travis green. It's pretty obvious your change is introducing any new breakages.

joeltine:
PR looks good. Can you please squash your commits into 1 and add a unit test to https://github.com/google/closure-library/blob/master/closure/goog/ui/container_test.js?

joeltine:
@ghosttie waiting on the last comment before merging.

ghosttie:
Sorry, I do intend to do squash the commits and add a unit test

ghosttie:
When I was writing the tests I realized it was working even without my changes.  highlightedIndex_ is set in goog.ui.Container.prototype.handleHighlightItem so it doesn't need to be set directly in setHighlightedIndex.  I can't think why I had the problem in the first place.

PR #737, opened 2016-07-21
Use mvn instead of ant to build the compiler

MatrixFrog:
I think this should fix the build...

### DIFF
--- a/scripts/ci/compile_closure.sh
+++ b/scripts/ci/compile_closure.sh
@@ -5,7 +5,7 @@
 # TODO(joeltine): Make strictMissingRequire an error when 
 # @suppress {missingRequire} works for it.
 
-java -Xmx1G -jar ../closure-compiler/build/compiler.jar \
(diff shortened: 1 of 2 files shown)


scripts/ci/compile_closure.sh
```diff
@@ -5,7 +5,7 @@
 # TODO(joeltine): Make strictMissingRequire an error when 
 # @suppress {missingRequire} works for it.
 
-java -Xmx1G -jar ../closure-compiler/build/compiler.jar \
+java -Xmx1G -jar ../closure-compiler/target/closure-compiler-1.0-SNAPSHOT.jar \
```
joeltine: What is 1.0-SNAPSHOT? Do we have to update this in the future?

MatrixFrog: This is because you're building from head. You can use the released compiler (from maven central) instead but when I set this up, I did this because
1. It was easier, and I was lazy
2. This lets us detect immediately if the compiler team lands a change that's going to break things in Closure, or if a PR in Closure is going to break when compiled with unreleased compiler changes.

PR #738, opened 2016-07-21
Update XPF and EUR currencies

DavidSkazy:
XPF currency was wrong => update XPF currency.
Most countries in Euro area are using the € in last position with a space => update EUR currency.
More details -> http://publications.europa.eu/code/fr/fr-370303.htm (French)

### DIFF
--- a/closure/goog/i18n/currency.js
+++ b/closure/goog/i18n/currency.js
@@ -288,7 +288,7 @@ goog.i18n.currency.CurrencyInfo = {
   'DOP': [2, 'RD$', 'RD$'],
   'EGP': [2, '£', 'LE'],
   'ETB': [2, 'Birr', 'Birr'],
-  'EUR': [2, '€', '€'],
+  'EUR': [50, '€', '€'],
   'GBP': [2, '£', 'GB£'],
   'HKD': [2, '$', 'HK$'],
   'HRK': [2, 'kn', 'kn'],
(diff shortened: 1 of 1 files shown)


joeltine:
@fabalbon, I don't see currency.js listed here: https://github.com/google/closure-library/wiki/Internationalization-(i18n)-changes-in-Closure-Library. Isn't it also auto-generated from CLDR?

fabalbon:
@joeltine currency.js is the only locale data file not being auto-generated. Still, we try be consistent with the data on CLDR.

joeltine:
Thanks @fabalbon.

@DavidSkazy, we're running global tests on this change internally to see if we break any existing users. @fabalbon should have an idea of the impact soon.

fabalbon:
Thanks for pointing out the bug in the **XPF** currency, we are updating it internally according to the data available on CLDR and it will be out on the next closure release. Regarding **EUR**, it has a broader impact thus we won't update at this moment.

We strongly discourage the use of the APIs on `currency.js` to get the the currency sign’s position, several of these functions have already been deprecated. This is because the position of the currency sign depends on the display locale, but the data on `goog.i18n.currency.CurrencyInfo` is based on the currency’s default locale, which is a problem with currencies such as **EUR** which don't have a default locale.

To format currencies we recommend the use of `goog.i18n.NumberFormat(goog.i18n.NumberFormat.Format.CURRENCY, “EUR")` which uses locale-dependent data which is auto-generated from CLDR.

joeltine:
I'm assuming this was fixed in the last release https://github.com/google/closure-library/releases

PR #739, opened 2016-07-29
Fix failed scan es6 destructuring assignment import

redforks:
depswriter.py support scans code like:

const {A, B} = goog.require('some.goog.module');

### DIFF
--- a/closure/bin/build/source.py
+++ b/closure/bin/build/source.py
@@ -23,12 +23,15 @@ __author__ = 'nnaze@google.com'
 
 import codecs
 import re
+import sys
 
 _BASE_REGEX_STRING = r'^\s*goog\.%s\(\s*[\'"](.+)[\'"]\s*\)'
 _MODULE_REGEX = re.compile(_BASE_REGEX_STRING % 'module')
 _PROVIDE_REGEX = re.compile(_BASE_REGEX_STRING % 'provide')
 
-_REQUIRE_REGEX_STRING = (r'^\s*(?:(?:var|let|const)\s+[a-zA-Z_$][a-zA-Z0-9$_]*'
+_SYMBOL_REGEX_STRING = r'[a-zA-Z_$][a-zA-Z0-9$_]*'
+_REQUIRE_REGEX_STRING = (r'^\s*(?:(?:var|let|const)\s+{?' + _SYMBOL_REGEX_STRING + 
+                         '(?:\s*,\s*' + _SYMBOL_REGEX_STRING + ')*}?'
                          r'\s*=\s*)?goog\.require\(\s*[\'"](.+)[\'"]\s*\)')
 _REQUIRES_REGEX = re.compile(_REQUIRE_REGEX_STRING)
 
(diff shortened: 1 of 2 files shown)


redforks:
Well, the continuous-integration test failed is not my commit fault. How do I re-run the check?

joeltine:
I restarted the CI build.

mashedcode:
This patch is somewhat important even though the closure-library itself does not make use of es6  because unit tests without compilation in browsers that do not support es6  would fail if the closure-library would start to use es6.
But apart from that others will use es6 in their project and it's very likely that they use the build tool provided by the closure-library.
The regex provided by @redforks is obviously not fully extensive since we can make use of much more complicated destructuring assignments in es6 but IMHO an acceptable trade-off if you do not want to use a regex such as `goog\.require\(\s*[\'"](.+)[\'"]\s*\)` (with no `^`).

closure/bin/build/source_test.py
```diff
@@ -57,7 +57,7 @@ def testSourceScanGoogModule(self):
 
     self.assertEqual(set(['foo']),
                      test_source.provides)
-    self.assertEqual(set(['bar']),
+    self.assertEqual(set(['bar', 'foo', 'foobar']),
```
joeltine: I think I'd prefer you move the ES6 module test to its own unit test method. Also, you can add a specific `_TEST_ES6_MODULE_SOURCE` for that test alone.

closure/bin/build/source.py
```diff
@@ -23,12 +23,15 @@
...
 import re
+import sys
 
 _BASE_REGEX_STRING = r'^\s*goog\.%s\(\s*[\'"](.+)[\'"]\s*\)'
 _MODULE_REGEX = re.compile(_BASE_REGEX_STRING % 'module')
 _PROVIDE_REGEX = re.compile(_BASE_REGEX_STRING % 'provide')
 
-_REQUIRE_REGEX_STRING = (r'^\s*(?:(?:var|let|const)\s+[a-zA-Z_$][a-zA-Z0-9$_]*'
+_SYMBOL_REGEX_STRING = r'[a-zA-Z_$][a-zA-Z0-9$_]*'
+_REQUIRE_REGEX_STRING = (r'^\s*(?:(?:var|let|const)\s+{?' + _SYMBOL_REGEX_STRING +
```
joeltine: we have an 80 col limit. The proper formatting is something like:

```
_SYMBOL_REGEX_STRING = r'[a-zA-Z_$][a-zA-Z0-9$_]*'
_REQUIRE_REGEX_STRING = (r'^\s*(?:(?:var|let|const)\s+{?' + _SYMBOL_REGEX_STRING
                         + '(?:\s*,\s*' + _SYMBOL_REGEX_STRING + ')*}?'
                         r'\s*=\s*)?goog\.require\(\s*[\'"](.+)[\'"]\s*\)')
```

joeltine:
Basic change LGTM, minus a couple comments. Can merge when they're addressed, and please remember to squash your commits to 1.

OOC, did you get this regex anywhere or did you just make it up yourself? I.e., I'm wondering if it came from a trusted/tested source as regex matching can be tricky.

joeltine:
Feel free to re-open once code review comments are addressed.

Issue #703, opened 2016-05-06
goog.math.Integer Multiplication Is Returning An Incorrect Value


xcthulhu:
I am currently getting the wrong value when I perform the following multiplication (using the google closure library version 0.0-20151016-61277aea):

```js
var b = goog.math.Integer.fromString(
        "f729d763a14ecd55ffffebab43f388d0f7cbae584d3765d509b5557d6048ea0c", 16);
console.log(b.multiply(b).toString(16))
```

this prints

eea1c478f4683b323f0953c9c8e067e2967d97e7ed0bf05862cecac60f30077f170e480beee2cd0c1d5516764d58bc260cafe5705bc6b6df63cf4c057cb9f090

The correct value can be computed with python:

```
b = 0xf729d763a14ecd55ffffebab43f388d0f7cbae584d3765d509b5557d6048ea0c
print hex(b*b)
```

this prints

0xeea1c478f4683b323f0953c9c8e067e3967d97e7ed0bf05862cecac60f30077f170e480beee2cd0c1d5516764d58bc260cafe5705bc6b6df63cf4c057cb9f090L

It is worth noting that the first 64 bits and the last 64 bits of the two printed values agree.

gfredericks:
The smallest example of this bug I could find is 

```
ffffffff00000001 · 1ffffffffffffffff
```

which should be

```
1fffffffe0000000100000000ffffffff
```

but `goog.math.Integer` gives:

```
1fffffffd0000000100000000ffffffff
```

concavelenz:
Thanks for the report

joeltine:
Is this no longer an issue?

gfredericks:
My pull request was merged, so I assume it's not an issue on master at
least. I haven't checked if it's been released.

joeltine:
Ah, sorry. Didn't even notice that.

PR #741, opened 2016-08-08
Fix multiplication carry in Integer

gfredericks:
Fixes issue #703

### DIFF
--- a/closure/goog/math/integer.js
+++ b/closure/goog/math/integer.js
@@ -566,6 +566,7 @@ goog.math.Integer.carry16_ = function(bits, index) {
   while ((bits[index] & 0xFFFF) != bits[index]) {
     bits[index + 1] += bits[index] >>> 16;
     bits[index] &= 0xFFFF;
+    index++;
   }
 };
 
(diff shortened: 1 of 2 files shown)


gfredericks:
I don't understand [this](https://travis-ci.org/google/closure-library/builds/150660466) build failure. It seems to have a compile error on completely unrelated code.

gfredericks:
I added an extra test for the smallest failing case I could find. Travis reports that the latest master build failed as well, so I'm not sure how to tell if I did something wrong. The integer tests pass on my machine.

xcthulhu:
I found the original bug with a more elaborate test involving modular square roots:

https://github.com/Sepia-Officinalis/secp256k1/blob/master/test/secp256k1/math_test.cljs#L22

@gfredericks We talked out of band about using this test, should I implement it here?

xcthulhu:
Similarly, modular multiplicative inverse is a good way of testing both multiplication and division algorithms are implemented - these would be sort of *math integration tests* but they are strong validations that everything is implemented properly.

PR #743, opened 2016-08-15
Fix bug: incorrect device pixel ratio when window.devicePixelRatio is not present

mashedcode:
When `window.devicePixelRatio` is not available the device pixel ratio is incorrect because of a bug that was never noticed because of an incorrect test case.

`min-resolution` will always match on a value lower than the actual resolution because the `min-resolution` rule will match when the browser **at least** provides the specified resolution. So the resolution needs to be tested top down instead of bottom up when using min-resolution in the media query.

Add IE10 support. IE10 does not support the dppx unit therefore, a dpi fallback is required.

Remove WebKit media query. WebKit always supports `window.devicePixelRatio` before `window.matchMedia` so the WebKit check is unnecessary.

Further note: It would be totally possible to create a fallback for FF3.5-5, IE9 and Opera versions before 12.1.

### DIFF
--- a/closure/goog/dom/dom.js
+++ b/closure/goog/dom/dom.js
@@ -2236,9 +2236,11 @@ goog.dom.getPixelRatio = function() {
   if (goog.isDef(win.devicePixelRatio)) {
     return win.devicePixelRatio;
   } else if (win.matchMedia) {
-    return goog.dom.matchesPixelRatio_(.75) ||
-        goog.dom.matchesPixelRatio_(1.5) || goog.dom.matchesPixelRatio_(2) ||
-        goog.dom.matchesPixelRatio_(3) || 1;
+    // Should be for IE10 and FF6-17 (this basically clamps to lower)
+    // Note that the order of these statements is important
+    return goog.dom.matchesPixelRatio_(3) || goog.dom.matchesPixelRatio_(2) ||
+           goog.dom.matchesPixelRatio_(1.5) || goog.dom.matchesPixelRatio_(1) ||
+           .75;
   }
   return 1;
 };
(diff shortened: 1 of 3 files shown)


mashedcode:
This looks good to me now. Please merge #745 for CI to pass.

mashedcode:
I might be a bit impatient but may I get some feedback?

mashedcode:
Proof of bug existence
---
```
// ==ClosureCompiler==
// @compilation_level ADVANCED_OPTIMIZATIONS
// @output_file_name default.js
// @use_closure_library true
// ==/ClosureCompiler==

goog.module('dpx');
const dom = goog.require('goog.dom');
// Obviously this throws a warning
const pixelRatio = dom.matchesPixelRatio_;
const dpx = pixelRatio(.75) || pixelRatio(1.5) || pixelRatio(2) || pixelRatio(3);
alert(dpx);
```
[Compile it!](https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540compilation_level%2520ADVANCED_OPTIMIZATIONS%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540use_closure_library%2520true%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250Agoog.module('dpx')%253B%250Aconst%2520dom%2520%253D%2520goog.require('goog.dom')%253B%250A%252F%252F%2520Obviously%2520this%2520throws%2520a%2520warning%250Aconst%2520pixelRatio%2520%253D%2520dom.matchesPixelRatio_%253B%250Aconst%2520dpx%2520%253D%2520pixelRatio(.75)%2520%257C%257C%2520pixelRatio(1.5)%2520%257C%257C%2520pixelRatio(2)%2520%257C%257C%2520pixelRatio(3)%253B%250Aalert(dpx)%253B)
[Try it!](https://jsfiddle.net/qcg7wup1/)

joeltine:
Sorry, we're woefully understaffed with too much to do atm. I can take a proper look at this on Tuesday.

closure/goog/dom/dom.js
```diff
@@ -2236,9 +2236,10 @@ goog.dom.getPixelRatio = function() {
   if (goog.isDef(win.devicePixelRatio)) {
     return win.devicePixelRatio;
   } else if (win.matchMedia) {
-    return goog.dom.matchesPixelRatio_(.75) ||
-        goog.dom.matchesPixelRatio_(1.5) || goog.dom.matchesPixelRatio_(2) ||
-        goog.dom.matchesPixelRatio_(3) || 1;
+    // Should be for IE10 and FF6-17 (this basically clamps to lower)
+    return goog.dom.matchesPixelRatio_(3) || goog.dom.matchesPixelRatio_(2) ||
```
joeltine: If ordering of these matters now, note it in the comment.

joeltine:
Seems reasonable enough to me. I'll need to run some tests on this PR internally to make sure it doesn't break existing tests relying on old behavior.

closure/goog/dom/dom.js
mashedcode: I added a comment and rebased.
Note that the previous author knew that the order would be important too.

joeltine:
You'll also need to squash your commits into 1 before I'll be able to merge this: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

mashedcode:
@joeltine I've done that.
I thought I was off pretty good with only three commits for the entire thing though.
It feels weird to squash an entire pull request into one commit since you could simply revert an entire merge too.
Anyways all good :smile:

joeltine:
Everything passed internally for this. Thanks for the change.

PR #745, opened 2016-08-15
Fix linter path

mashedcode:
Since the closure compiler moved from ant to maven the path to the linter has to be changed too.

This may be considered related to 8b3e2795d3b2d40d67eacc2bfa02c30718eda04c.


joeltine:
Huh, any clue why build on master is still passing even without this change?

joeltine:
Oh... it is broken. It just hasn't run in a while. Thanks.

PR #757, opened 2016-09-16
Update deprecated KeyEvent keyIdentifier call

afathman:
https://www.chromestatus.com/features/5316065118650368 keyIdentifier will no longer be supported as of Chrome 54.

### DIFF
--- a/closure/goog/events/keyhandler.js
+++ b/closure/goog/events/keyhandler.js
@@ -429,11 +429,11 @@ goog.events.KeyHandler.prototype.handleEvent = function(e) {
       }
     }
   } else if (
-      be.keyIdentifier &&
-      be.keyIdentifier in goog.events.KeyHandler.keyIdentifier_) {
+      be.key &&
+      be.key in goog.events.KeyHandler.keyIdentifier_) {
     // This is needed for Safari Windows because it currently doesn't give a
     // keyCode/which for non printable keys.
-    key = goog.events.KeyHandler.keyIdentifier_[be.keyIdentifier];
+    key = goog.events.KeyHandler.keyIdentifier_[be.key];
   }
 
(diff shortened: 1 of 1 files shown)


afathman:
I signed the agreement! 👍

mashedcode:
@afathman I figured that you can rerun the CI test yourself by force pushing with no modification.

closure/goog/events/keyhandler.js
```diff
@@ -429,11 +429,11 @@ goog.events.KeyHandler.prototype.handleEvent = function(e) {
       }
     }
   } else if (
-      be.keyIdentifier &&
-      be.keyIdentifier in goog.events.KeyHandler.keyIdentifier_) {
+      be.key &&
+      be.key in goog.events.KeyHandler.keyIdentifier_) {
```
mashedcode: Use clang-format. These three lines fit on one line now.

closure/goog/events/keyhandler.js
```diff
@@ -429,11 +429,11 @@ goog.events.KeyHandler.prototype.handleEvent = function(e) {
       }
     }
   } else if (
-      be.keyIdentifier &&
-      be.keyIdentifier in goog.events.KeyHandler.keyIdentifier_) {
+      be.key &&
```
mashedcode: I'm not highly familiar with this code but this seems to cover an edge-case. Maybe the use of the deprecated `keyIdentifier` property is justified?
Please explain why you think that this is an issue.
Furthermore please add a test case for this specific edge-case and explain when this code will be reached.
Note that `keyIdentifier` is still being used in test code: https://github.com/google/closure-library/blob/8bbc453a8808aa77af91f959fb0dc97be058cd97/closure/goog/events/keyhandler_test.js#L705

mashedcode:
Please close this PR.

PR #758, opened 2016-09-20
Export _displayStringForValue for mockmatchers

KamranAsif:
[goog.testing.mockmatters](https://github.com/google/closure- library/blob/master/closure/goog/testing/mockmatchers.js) relies on _displayStringForValue from goog.testing.asserts. However, that function is not exported from goog.testing.asserts and breaks UMD.

I'm not sure if there is a better approach here, since _displayStringForValue seems to indicate a private function

### DIFF
--- a/closure/goog/testing/asserts.js
+++ b/closure/goog/testing/asserts.js
@@ -1338,3 +1338,4 @@ goog.exportSymbol('assertRoughlyEquals', assertRoughlyEquals);
 goog.exportSymbol('assertContains', assertContains);
 goog.exportSymbol('assertNotContains', assertNotContains);
 goog.exportSymbol('assertRegExp', assertRegExp);
+goog.exportSymbol('_displayStringForValue', _displayStringForValue);


KamranAsif:
I signed it!

joeltine:
Can you give me detail about exactly how you're encountering this error? In general, this should only be an issue if you're compiling your tests using mockmatchers. The mockmatchers library is strictly not a "compilation friendly" class (it has other issues outside of these exported symbols). 

In general the use of the private `_displayStringForValue()` in mockmatchers seems like a terrible hack. Exporting it just makes it worse. If anything, I would just fix the use of  that method to use something local in mockmatchers.js.

KamranAsif:
We are using a modified closure-loader for our unit tests using intern. We had to add certain hacks to export globals such as asserts, but since this function isn't exported we couldn't grab it in the global scope.

I agree that the private usage is pretty bad, however it depends on _trueTypeOf which is a super long function that would also have to be copied over. 

Any ideas for a better approach?

joeltine:
> Any ideas for a better approach?

The best thing would be to extract the method(s) into a shared, public, namespace and use them between the two files (asserts.js and mockmatchers.js). Believe me when I say the way this is used right now is pretty cringeworthy. I don't like it all. Unless you're willing to do the legwork to extract out the shared code (which might be tricky), I'd suggest continuing to work around it with your own hack.

PR #765, opened 2016-10-20
Allow for custom timezone offsets

austinfathman:
Allows for custom timezone offsets in favor over
client side timezone for Today button and italicized date
If value is set to null (default), calendar operates normally

Use case: User has preferred a timezone and wishes calendar to behave as if the user were in the preferred timezone.

### DIFF
--- a/closure/goog/ui/datepicker.js
+++ b/closure/goog/ui/datepicker.js
@@ -28,6 +28,7 @@ goog.require('goog.asserts');
 goog.require('goog.date.Date');
 goog.require('goog.date.DateRange');
 goog.require('goog.date.Interval');
+goog.require('goog.date.UtcDateTime');
 goog.require('goog.dom');
 goog.require('goog.dom.NodeType');
 goog.require('goog.dom.TagName');
@@ -297,6 +298,14 @@ goog.ui.DatePicker.prototype.elMonthYear_ = null;
  */
 goog.ui.DatePicker.prototype.elFootRow_ = null;
 
+/**
+ * The offset from GMT/UTC time in minutes.
+ * Example: EDT would be 240 and JST -540.
+ * @type {number}
+ * @private
+ */
+goog.ui.DatePicker.prototype.timezoneOffset_ = null;
+
 
 /**
  * Generator for unique table cell IDs.
@@ -343,6 +352,33 @@ goog.ui.DatePicker.Events = {
 goog.ui.DatePicker.prototype.isCreated =
     goog.ui.DatePicker.prototype.isInDocument;
 
+/**
+ * @return {number} The timezone offset from UTC in minutes.
+ */
+goog.ui.DatePicker.prototype.getTimezoneOffset = function() {
+  return this.timezoneOffset_;
+};
+
+/**
+ * @return {number} The timezone offset from UTC in milliseconds.
+ * @private
+ */
+goog.ui.DatePicker.prototype.getTimezoneOffsetMs_ = function() {
+  return this.timezoneOffset_ * 60 * 1000;
+};
+
+/**
+* @return {goog.date.Date} The Date object for today.
+* @private
+*/
+goog.ui.DatePicker.prototype.getToday_ = function() {
+  if (goog.isNull(this.timezoneOffset_)) {
+    return new goog.date.Date();
+  }
+  var date = new goog.date.UtcDateTime();
+  date.setTime(date.getTime() - this.getTimezoneOffsetMs_());
+  return date;
+};
 
 /**
  * @return {number} The first day of week, 0 = Monday, 6 = Sunday.
@@ -431,6 +467,15 @@ goog.ui.DatePicker.prototype.getBaseCssClass = function() {
   return goog.ui.DatePicker.BASE_CSS_CLASS_;
 };
 
+
+/**
+ * Sets the timezone offset.
+ *
+ * @param {number} offset The timezone offset from UTC in minutes.
+ */
+goog.ui.DatePicker.prototype.setTimezoneOffset = function(offset) {
+  this.timezoneOffset_ = offset;
+};
 
 /**
  * Sets the first day of week
@@ -689,7 +734,7 @@ goog.ui.DatePicker.prototype.nextYear = function() {
  * Selects the current date.
  */
 goog.ui.DatePicker.prototype.selectToday = function() {
-  this.setDate(new goog.date.Date());
+  this.setDate(this.getToday_());
 };
 
 
@@ -1423,7 +1468,7 @@ goog.ui.DatePicker.prototype.redrawCalendarGrid_ = function() {
   }
 
   var month = this.activeMonth_.getMonth();
-  var today = new goog.date.Date();
+  var today = this.getToday_();
   var todayYear = today.getFullYear();
   var todayMonth = today.getMonth();
   var todayDate = today.getDate();


austinfathman:
I submitted!

afathman:
I authored these commits

austinfathman:
I authored this commit.

mashedcode:
It still seems weird to me that you don't want to trust the timezone set on the client. If the user prefers a different timezone he should usually change the timezone on his device.
There might obviously be a permission issue with setting the timezone on the device but in that case the issue is probably greater because if the timezone on the device is incorrect how can one be sure that the time is correct.

closure/goog/ui/datepicker.js
```diff
@@ -1423,7 +1464,13 @@ goog.ui.DatePicker.prototype.redrawCalendarGrid_ = function() {
   }
 
   var month = this.activeMonth_.getMonth();
-  var today = new goog.date.Date();
+  var today;
```
austinfathman: Yea, I hate redundant code too. I'll abstract this out to a function and add it in my next commit.

closure/goog/ui/datepicker.js
```diff
@@ -689,7 +723,14 @@ goog.ui.DatePicker.prototype.nextYear = function() {
  * Selects the current date.
  */
 goog.ui.DatePicker.prototype.selectToday = function() {
-  this.setDate(new goog.date.Date());
+  var date;
+  if (this.timezoneOffset_) {
```
austinfathman: Cool function, will use from now on!

closure/goog/ui/datepicker.js
```diff
@@ -433,6 +458,15 @@ goog.ui.DatePicker.prototype.getBaseCssClass = function() {
 
 
 /**
+ * Sets the timezone offset
+ *
+ * @param {number} the timezone offset in hours from UTC
```
austinfathman: Done, thanks!

closure/goog/ui/datepicker.js
```diff
@@ -343,6 +353,21 @@ goog.ui.DatePicker.Events = {
...
+ */
+goog.ui.DatePicker.prototype.getTimezoneOffset = function() {
+  return this.timezoneOffset_;
+};
+
+/**
+ * @return {number} The timezone hour offset from UTC in milliseconds
+ * @private
+ */
+goog.ui.DatePicker.prototype.getTimezoneOffsetMS_ = function() {
```
austinfathman: Good catch, thanks

closure/goog/ui/datepicker.js
```diff
@@ -343,6 +353,21 @@ goog.ui.DatePicker.Events = {
 goog.ui.DatePicker.prototype.isCreated =
     goog.ui.DatePicker.prototype.isInDocument;
 
+/**
+ * @return {number} The timezone hour offset from UTC
```
austinfathman: Added a period to all of the sentences.

austinfathman:
@mashedcode, thanks for looking over my PR. I'll look into re-generating deps.js in the morning as well as getting some test cases setup, however I've addressed the rest of your line comments.

svieira:
@mashedcode - the reason we need this is because we're not using the device's date and time but rather the user's preferred time zone (which might not be the same timezone as the time zone that they are currently in / their device is set to).

austinfathman:
Made a quick modification: changed the offset from hours to minutes to avoid using decimal numbers / match Javascript's internal Date.getTimezoneOffset and popular timezone libraries (eg moment-timezone.js).

closure/goog/ui/datepicker.js
```diff
@@ -343,6 +353,34 @@ goog.ui.DatePicker.Events = {
...
+*/
+goog.ui.DatePicker.prototype.getToday_ = function() {
+  if (goog.isNull(this.timezoneOffset_)) {
+    return new goog.date.Date();
+  }
+  var date = new goog.date.UtcDateTime();
+  date.setTime(date.getTime() - this.getTimezoneOffsetMs_());
+  return date;
+};
+
```
roseswang: (nit) extra line

closure/goog/ui/datepicker.js
```diff
@@ -297,6 +298,15 @@ goog.ui.DatePicker.prototype.elMonthYear_ = null;
...
 
+/**
+ * The offset from GMT/UTC time in minutes.
+ * Example: EDT would be 240 and JST -540.
+ * @type {number}
+ * @private
+ */
+goog.ui.DatePicker.prototype.timezoneOffset_ = null;
+
+
```
austinfathman: Thanks, fixed

closure/goog/ui/datepicker.js
```diff
@@ -343,6 +352,33 @@ goog.ui.DatePicker.Events = {
...
+/**
+* @return {goog.date.Date} The Date object for today.
+* @private
+*/
+goog.ui.DatePicker.prototype.getToday_ = function() {
+  if (goog.isNull(this.timezoneOffset_)) {
+    return new goog.date.Date();
+  }
+  var date = new goog.date.UtcDateTime();
```
mashedcode: Simply use `Date` instead of `goog.date.UtcDateTime`  because you don't use any of the features that `goog.date.UtcDateTime` provides to you and when `goog.ui.DatePicker.prototype.setDate` will be called it'll build a new `goog.date.Date` anyways and furthermore it does already accept a `Date` too.
Also since `goog.date.Date` mimics `Date` so this won't be an issue in `goog.ui.DatePicker.prototype.redrawCalendarGrid_` either.
The only reason you can't use `goog.date.Date` directly is that it strips the time information.

closure/goog/ui/datepicker.js
```diff
@@ -343,6 +352,33 @@ goog.ui.DatePicker.Events = {
...
+/**
+ * @return {number} The timezone offset from UTC in milliseconds.
+ * @private
+ */
+goog.ui.DatePicker.prototype.getTimezoneOffsetMs_ = function() {
+  return this.timezoneOffset_ * 60 * 1000;
+};
+
+/**
+* @return {goog.date.Date} The Date object for today.
```
mashedcode: Date -> date

closure/goog/ui/datepicker.js
```diff
@@ -297,6 +298,14 @@ goog.ui.DatePicker.prototype.elMonthYear_ = null;
  */
 goog.ui.DatePicker.prototype.elFootRow_ = null;
 
+/**
+ * The offset from GMT/UTC time in minutes.
+ * Example: EDT would be 240 and JST -540.
+ * @type {number}
```
mashedcode: This may be null so it should be `@type {?number}`.

austinfathman:
@mash, you're exactly right about opt_date also needing to be handled if it's null, as it would display the wrong date if it was not set and Timezone offset was. I'll address the rest of your CR comments tomorrow/monday.

closure/goog/ui/datepicker.js
joeltine: +1

closure/goog/ui/datepicker.js
```diff
@@ -343,6 +352,33 @@ goog.ui.DatePicker.Events = {
 goog.ui.DatePicker.prototype.isCreated =
     goog.ui.DatePicker.prototype.isInDocument;
 
+/**
+ * @return {number} The timezone offset from UTC in minutes.
```
joeltine: ?number

closure/goog/ui/datepicker.js
```diff
@@ -343,6 +352,33 @@ goog.ui.DatePicker.Events = {
...
+goog.ui.DatePicker.prototype.getTimezoneOffset = function() {
+  return this.timezoneOffset_;
+};
+
+/**
+ * @return {number} The timezone offset from UTC in milliseconds.
+ * @private
+ */
+goog.ui.DatePicker.prototype.getTimezoneOffsetMs_ = function() {
+  return this.timezoneOffset_ * 60 * 1000;
```
joeltine: Need to account for null on `timezoneOffset`

closure/goog/ui/datepicker.js
joeltine: `{!goog.date.Date}`

closure/goog/ui/datepicker.js
```diff
@@ -433,6 +469,15 @@ goog.ui.DatePicker.prototype.getBaseCssClass = function() {
 
 
 /**
+ * Sets the timezone offset.
+ *
+ * @param {number} offset The timezone offset from UTC in minutes.
+ */
+goog.ui.DatePicker.prototype.setTimezoneOffset = function(offset) {
+  this.timezoneOffset_ = offset;
```
joeltine: I think to complete this change, you'd also need to update `this.date_` here. Otherwise other methods that rely on `date_` will be incorrect. E.g., calling `getDate()` immediately after setting an offset will not return the expected time.

joeltine:
Conceptually, the change seems fine to me. A few concerns:

1. You need to add unit tests. See ui/datepicker_test.js.
2. goog.ui is in maintenance mode and we're generally only accepting major bug fixes. To land this, you'd need someone to champion it for you and get it submitted internally. I will try to do this but if any troubles arise that require non-trivial work, I'll have to pass on it.

PR #768, opened 2016-11-08
Teach travis how to regenerate gh-pages.

shicks:
Adds a secure token and a script to generate gh-pages.  I've run the script manually (passing in what I believe Travis will pass for the various environment variables) and confirmed that it works.  Next step is to push to master and see if that side is correct.

### DIFF
--- a/scripts/ci/generate_latest_docs.sh
+++ b/scripts/ci/generate_latest_docs.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+# Rebuild gh-pages branch, adapted from
+# https://github.com/google/dagger/blob/master/util/generate-latest-docs.sh
+
+DIR=$(mktemp -d)
+mkdir -p "$DIR"
+
+if [ "$TRAVIS_REPO_SLUG" = "google/closure-library" -a \
+     "$TRAVIS_PULL_REQUEST" = "false" -a \
+     "$TRAVIS_BRANCH" = "master" ]; then
+  echo -e "Publishing documentation...\n"
+
+  # Files to omit from documentation
+  declare -A BLACKLIST
+  BLACKLIST['events/eventtargettester.js']=true
+  BLACKLIST['i18n/compactnumberformatsymbolsext.js']=true
+  BLACKLIST['i18n/datetimepatternsext.js']=true
+  BLACKLIST['i18n/numberformatsymbolsext.js']=true
+  BLACKLIST['i18n/listsymbolsext.js']=true
+  BLACKLIST['promise/testsuiteadapter.js']=true
+  BLACKLIST['proto2/package_test.pb.js']=true
+  BLACKLIST['proto2/test.pb.js']=true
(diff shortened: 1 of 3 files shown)


scripts/ci/generate_latest_docs.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+
+# Rebuild gh-pages branch, adapted from
+# https://github.com/google/dagger/blob/master/util/generate-latest-docs.sh
+
+HOME=/tmp/aaa
```
joeltine: `$(mktmp)`?

scripts/ci/generate_latest_docs.sh
```diff
@@ -0,0 +1,86 @@
...
+    --type_filter '^.*+_$'
+  )
+
+  while read -r file; do
+    file=${file#./}
+    if [[ "$file" =~ ^closure/goog/demos ]]; then continue; fi
+    if [[ "$file" =~ _test\.js$ || "$file" =~ _perf\.js$ ]]; then continue; fi
+    if [[ "${BLACKLIST[${file#closure/goog/}]}" == true ]]; then continue; fi
+    if [[ "$file" =~ ^closure/goog || "$file" =~ ^third_party/closure/goog ]]; then
+      command=("${command[@]}" --source "$file")
```
joeltine: You really have to run the command on every single file? I thought dossier would just operate on a root directory.

scripts/ci/generate_latest_docs.sh
```diff
@@ -0,0 +1,86 @@
...
+  done < <(find . -name '*.js')
+
+  # Run dossier.
+  "${command[@]}"
+
+  # Make a commit and push it.
+  (
+    cd $HOME/gh-pages
+    git add -A
+    git commit -m "Latest documentation on successful travis build $TRAVIS_BUILD_NUMBER auto-pushed to gh-pages"
```
joeltine: Include `$TRAVIS_COMMIT` too.

concavelenz:
Nice

scripts/ci/generate_latest_docs.sh
shicks: Oops, that was just so I could test it on my local machine without dumping lots of crap directly into my home directory.  Might make sense just to do the whole thing in tmp regardless.

scripts/ci/generate_latest_docs.sh
shicks: It does, but we're filtering out various files in a way that is more complicated than what dossier can do by itself, presumably.

This is basically just adapted from the python script we use internally, complete with the filters and blacklist, etc.

scripts/ci/generate_latest_docs.sh
shicks: Done. Reworded slightly.

shicks:
Thanks for the review, merging.

PR #772, opened 2016-11-28
Set sudo: required on Travis config

joeltine:
Testing if this fixes our maven jscomp build flakiness


joeltine:
It seems like the maven build passes with this, so we'll give it a try. The failures are just sauce labs flakes unrelated to the jscomp build.

PR #774, opened 2016-12-01
Keep trying to fix generate_latest_docs.sh

shicks:
In particular, this fixes the current error (wrong java version on travis - needs to be Java8) and also fixes a few other problems with the script.  It now runs on all presubmit checks, but only prepares the commit, rather than actually deploying it, for presubmits.


concavelenz:
LGTM

PR #778, opened 2016-12-07
Fix Deferred addFinally errback scope

drauggres:
Errback in finally called in Deferred scope instead of opt_scope.
#334


drauggres:
I signed it!

shicks:
I submitted this internally but ended up having to roll it back due to conformance violations.  It might be a while before we can actually land it permanently.

shicks:
Closed by a53790dfb23e3014f2783b664d0058a4272ecdde

PR #782, opened 2016-12-19
Set unusedLocals as warning, for now

joeltine:
Recent compiler release introduced a bunch of new unused local failures


joeltine:
@MatrixFrog FYI

PR #788, opened 2017-01-17
Fix parsing empty cookies in IE/Edge

pauldraper:
IE/Edge accepts an empty Set-Cookie header and adds an empty part to document.cookie

I don't care how this aberrant cookie is parsed, but currently the parsing stops when it encounters this empty cookie.

### DIFF
--- a/closure/goog/net/cookies_test.js
+++ b/closure/goog/net/cookies_test.js
@@ -209,6 +209,16 @@ function testGetEmptyCookieIE() {
   assertEquals('', cookies.get('test3'));
 }
 
+function testGetReallyEmptyCookieIE() {
+  stubs.set(
+      cookies, 'getCookie_', function() { return 'test1; ; test3'; });
+
+  assertEquals('', cookies.get('test1'));
+  assertEquals('', cookies.get(''));
+  assertEquals('', cookies.get('test3'));
+  assertEquals(3, cookies.getCount());
+}
+
 // TODO(chrisn): Testing max age > 0 requires a mock clock.
 
 function mockSetCookie(var_args) {
--- a/closure/goog/net/cookies.js
+++ b/closure/goog/net/cookies.js
@@ -189,7 +189,8 @@ goog.net.Cookies.prototype.set = function(
 goog.net.Cookies.prototype.get = function(name, opt_default) {
   var nameEq = name + '=';
   var parts = this.getParts_();
-  for (var i = 0, part; part = parts[i]; i++) {
+  for (var i = 0, part; i < parts.length; i++) {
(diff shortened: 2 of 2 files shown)


mashedcode:
In IE8 and earlier, empty-string matches are ignored by `.split(RexExp)`. Therefore `getParts_` does not return the empty cookie in the first place and the test fails in IE8.

We could:

1. Ignore empty cookies in general.
2. Fix `getParts_` for lte IE8 to support empty cookies.

I'd prefer option one because the best use case for empty cookies that comes into my mind does not justify their existence if they're not supported by everyone.

closure/goog/net/cookies_test.js
```diff
@@ -209,6 +209,16 @@ function testGetEmptyCookieIE() {
   assertEquals('', cookies.get('test3'));
 }
 
+function testGetReallyEmptyCookieIE() {
+  stubs.set(
+      cookies, 'getCookie_', function() { return 'test1; ; test3'; });
```
mashedcode: Please [format JavaScript with clang-format](https://github.com/google/closure-library/wiki/Formatting-.js-with-clang-format).

pauldraper:
> I'd prefer option one because the best use case for empty cookies that comes into my mind does not justify their existence if they're not supported by everyone.

Agreed.

shicks:
This test is currently failing in IE8:

```
ERROR in testGetReallyEmptyCookieIE
Expected <> (String) but was <undefined>
```

mashedcode:
@shicks I've already explained why it fails [#788 ie8 comment](https://github.com/google/closure-library/pull/788#issuecomment-273505364).

shicks:
Sorry, I assumed since the author responded in agreement, that he implemented your suggestion.  I guess that wasn't the case.  We can't merge this until IE8 passes, so if we can get that test fixed, that would be great.

For your option 1 are you suggesting we simply or with undefined so that get() never returns empty string (or vice versa - I'd personally prefer never returning undefined)?  I don't know under what situations there's currently a difference (it sounds like newer browsers are capable of distinguishing empty strings from absent) so if we made that change I'd need to run it through some more rigorous internal testing to make sure we're not breaking anybody who's depending on one behavior or the other.

mashedcode:
@shicks no the suggestion of option 1 was to simply `if (part === '') continue;`.

shicks:
Resolved by using String.split(';') instead of the regex, and then goog.string.trim on the result.  Cookies should be read infrequently enough that this shouldn't be particularly problematic.

shicks:
Oops, closed the wrong PR with that last commit.

PR #797, opened 2017-02-17
Enable XhrIO.send to set XmlHttpFactory as optional paraneter

code-4-fun:
There is a use case when I want to use `XhrIO.send(....)` method. But current implementation does not allow us for use custom XmlHttpFactory. This pull request adds another optional parameter to `XhrIo.send(...)` method which accepts `XmlHttpFactory` instance as input and generates appropriate `xhr` instance.


code-4-fun:
I signed it!

code-4-fun:
I have also included committer email id to the agreement.

mashedcode:
I don't like this. Can't you just set the factory via [`goog.net.XmlHttp.setGlobalFactory`](https://github.com/google/closure-library/blob/ad4c2517dbc061d3de4eecf575b52c086dafe1d2/closure/goog/net/xmlhttp.js#L151)?

code-4-fun:
Hi @mashedcode, I would like to use `XhrIO.send` to post some data to my API's. However, I think it will be cool, to have the ability to use `XmlHttpFactory` as an option. This way, developers can avoid maintaining global factory instance. It will also be useful when one wants to use static send method for a short lived API call and has the ability to use his / her own implementation of `XmlHttpRequest`. Please let me know if this makes sense... :)

closure/goog/net/xhrio.js
```diff
@@ -332,12 +332,21 @@ goog.net.XhrIo.sendInstances_ = [];
...
+ *     creating XMLHttpRequest objects
  * @return {!goog.net.XhrIo} The sent XhrIo.
  */
 goog.net.XhrIo.send = function(
     url, opt_callback, opt_method, opt_content, opt_headers,
-    opt_timeoutInterval, opt_withCredentials) {
-  var x = new goog.net.XhrIo();
+    opt_timeoutInterval, opt_withCredentials, opt_xmlHttpFactory) {
+
+  var x;
```
mashedcode: I'd prefer the ternary operator instead of the `if` to avoid the uninitialized variable. See [5.1.3 Declared when needed, initialized as soon as possible](https://google.github.io/styleguide/jsguide.html#features-declared-when-needed) or [Deferred initialization](https://google.github.io/styleguide/javascriptguide.xml?showone=Deferred_initialization#Deferred_initialization).

closure/goog/testing/net/xhrio.js
```diff
@@ -125,12 +134,19 @@ goog.testing.net.XhrIo.cleanup = function() {
...
+ * @param {goog.net.XmlHttpFactory=} opt_xmlHttpFactory Factory to use when
+ *     creating XMLHttpRequest objects
  * @return {!goog.testing.net.XhrIo} The mocked sent XhrIo.
  */
 goog.testing.net.XhrIo.send = function(
     url, opt_callback, opt_method, opt_content, opt_headers,
-    opt_timeoutInterval, opt_withCredentials) {
-  var x = new goog.testing.net.XhrIo();
+    opt_timeoutInterval, opt_withCredentials, opt_xmlHttpFactory) {
+  var x;
```
mashedcode: I'd prefer the ternary operator instead of the `if` to avoid the uninitialized variable. See [5.1.3 Declared when needed, initialized as soon as possible](https://google.github.io/styleguide/jsguide.html#features-declared-when-needed) or [Deferred initialization](https://google.github.io/styleguide/javascriptguide.xml?showone=Deferred_initialization#Deferred_initialization).

closure/goog/testing/net/xhrio.js
```diff
@@ -125,12 +134,19 @@ goog.testing.net.XhrIo.cleanup = function() {
...
+  if (opt_xmlHttpFactory) {
+    x = new goog.testing.net.XhrIo(undefined, opt_xmlHttpFactory);
```
mashedcode: In my opinion `null` fits better than `undefined` in this case.

closure/goog/testing/net/xhrio_test.js
```diff
@@ -55,6 +55,44 @@ function testStaticSend() {
   assertEquals('sendInstances_ after READY', 0, sendInstances.length);
 }
 
+function testStaticSendWithFactory() {
+  var xhrFactoryCalled = 0, optionsFactoryCalled = 0, callbackInvoked = 0;
```
mashedcode: Please use [5.1.2 One variable per declaration](https://google.github.io/styleguide/jsguide.html#features-one-variable-per-declaration) even though the closure-library seems to conform to the old JavaScript style guide.

closure/goog/testing/net/xhrio_test.js
```diff
@@ -55,6 +55,44 @@ function testStaticSend() {
...
+        return goog.net.XmlHttp();
+      },
+      function() {
+        optionsFactoryCalled++;
+        return {};
+      });
+
+  sendInstances = goog.testing.net.XhrIo.getSendInstances();
+  var returnedXhr = goog.testing.net.XhrIo.send(
+      'url', undefined, 'GET', '', undefined, 0, false, wrapperFactory);
```
mashedcode: I'd prefer `null` instead of `undefined` see also a [call to goog.net.XhrIo.send](https://github.com/google/closure-library/blob/e877b1eac410c0d842bcda118689759512e0e26f/closure/goog/debug/errorreporter.js#L221).

closure/goog/testing/net/xhrio_test.js
```diff
@@ -55,6 +55,44 @@ function testStaticSend() {
...
+  assertEquals('sendInstances_ after send', 1, sendInstances.length);
+  xhr = sendInstances[sendInstances.length - 1];
+  assertTrue('isActive after request', xhr.isActive());
+  assertEquals(returnedXhr, xhr);
```
mashedcode: Missing message.

closure/goog/testing/net/xhrio_test.js
```diff
@@ -55,6 +55,44 @@ function testStaticSend() {
...
+        optionsFactoryCalled++;
+        return {};
+      });
+
+  sendInstances = goog.testing.net.XhrIo.getSendInstances();
+  var returnedXhr = goog.testing.net.XhrIo.send(
+      'url', undefined, 'GET', '', undefined, 0, false, wrapperFactory);
+  assertEquals('sendInstances_ after send', 1, sendInstances.length);
+  xhr = sendInstances[sendInstances.length - 1];
+  assertTrue('isActive after request', xhr.isActive());
```
mashedcode: Why does this need to be tested again if it was already tested in `testStaticSend`. It seems to me that most of the tests here are not affected by the custom factory.

closure/goog/testing/net/xhrio.js
```diff
@@ -453,6 +469,13 @@ goog.testing.net.XhrIo.prototype.send = function(
   if (this.testQueue_) {
     this.testQueue_.enqueue(['s', url, opt_method, opt_content, opt_headers]);
   }
+
+  // not interested in what factory does to create object. Contention is simply
+  // to create an instance.
+  this.createXhr();
+  this.xmlHttpFactory_ ? this.xmlHttpFactory_.getOptions() :
```
mashedcode: This should be an if statement since the result is not used.

closure/goog/testing/net/xhrio_test.js
```diff
@@ -55,6 +55,44 @@ function testStaticSend() {
...
+  xhr.simulateResponse(200, '');
+  assertFalse('isActive after response', xhr.isActive());
+  assertEquals(
+      'readyState after response', goog.net.XmlHttp.ReadyState.COMPLETE,
+      xhr.getReadyState());
+
+  xhr.simulateReady();
+  assertEquals('sendInstances_ after READY', 0, sendInstances.length);
+
+  // check if all wrapperFactory is used during send
```
mashedcode: [Complete sentences should use appropriate capitalization and punctuation.](https://google.github.io/styleguide/javascriptguide.xml?showone=Comments#Comments)

closure/goog/testing/net/xhrio.js
```diff
@@ -453,6 +469,13 @@ goog.testing.net.XhrIo.prototype.send = function(
   if (this.testQueue_) {
     this.testQueue_.enqueue(['s', url, opt_method, opt_content, opt_headers]);
   }
+
+  // not interested in what factory does to create object. Contention is simply
```
mashedcode: [Complete sentences should use appropriate capitalization.](https://google.github.io/styleguide/javascriptguide.xml?showone=Comments#Comments)

closure/goog/net/xhrio.js
```diff
@@ -530,8 +539,9 @@ goog.net.XhrIo.prototype.send = function(
 
   // Set up upload/download progress events, if progress events are supported.
   if (this.getProgressEventsEnabled() && 'onprogress' in this.xhr_) {
-    this.xhr_.onprogress =
-        goog.bind(function(e) { this.onProgressHandler_(e, true); }, this);
+    this.xhr_.onprogress = goog.bind(function(e) {
```
mashedcode: My clang-format says the previous revision is the correct version.

closure/goog/net/xhrio.js
```diff
@@ -563,8 +573,9 @@ goog.net.XhrIo.prototype.send = function(
 
   // Add headers specific to this request
   if (opt_headers) {
-    goog.structs.forEach(
-        opt_headers, function(value, key) { headers.set(key, value); });
+    goog.structs.forEach(opt_headers, function(value, key) {
```
mashedcode: My clang-format says the previous revision is the correct version.

closure/goog/net/xhrio.js
code-4-fun: fixed. This condition itself was not required as the argument for constructor itself is optional. Removed `if...else` block.

closure/goog/net/xhrio.js
code-4-fun: fixed

closure/goog/net/xhrio.js
code-4-fun: fixed.

closure/goog/testing/net/xhrio_test.js
code-4-fun: fixed

closure/goog/testing/net/xhrio_test.js
code-4-fun: fixed

closure/goog/testing/net/xhrio_test.js
code-4-fun: fixed

closure/goog/testing/net/xhrio_test.js
code-4-fun: removed these repeated test scenarios

closure/goog/testing/net/xhrio_test.js
code-4-fun: done

closure/goog/testing/net/xhrio.js
code-4-fun: done

closure/goog/testing/net/xhrio.js
code-4-fun: done

closure/goog/testing/net/xhrio.js
code-4-fun: done

closure/goog/testing/net/xhrio.js
code-4-fun: fixed. This condition is not required as the argument for constructor itself is optional. Removed `if...else` block.

closure/goog/net/xhrio.js
mashedcode: Seems unchanged to me.

mashedcode: Na your clang-format seems to be way off. See also CI. Read [Formatting .js with clang-format](https://github.com/google/closure-library/wiki/Formatting-.js-with-clang-format).

closure/goog/testing/net/xhrio_test.js
```diff
@@ -55,6 +55,28 @@ function testStaticSend() {
...
 }
 
+function testStaticSendWithFactory() {
+  var xhrFactoryCalled = 0;
+  var optionsFactoryCalled = 0;
+
+  var wrapperFactory = new goog.net.WrapperXmlHttpFactory(
+      function() {
+        xhrFactoryCalled++;
+        return goog.net.XmlHttp();
```
mashedcode: The use of `goog.net.XmlHttp` is incorrect here since this is a only the mock implementation of `goog.net.XhrIo`.
Historically it was the case that [`goog.net.XmlHttp` used to be the same as `MockXmlHttp`](https://github.com/google/closure-library/blob/dfe700923ca161b600cd7c057ece702841c18d51/closure/goog/net/xhrio_test.html#L81) in `goog.net.XhrIoTest` today.
Probably `MockXmlHttp` should be moved somewhere where `goog.testing.net.XhrIoTest` can use it too but since `goog.net.XmlHttp` is not globally mangled by the `MockXmlHttp` alias anymore this is all incorrect now and you either want to remove the use of `goog.net.XmlHttp();` or refactor as mentioned above.

closure/goog/testing/net/xhrio.js
```diff
@@ -467,7 +487,8 @@ goog.testing.net.XhrIo.prototype.send = function(
  * @protected
  */
 goog.testing.net.XhrIo.prototype.createXhr = function() {
-  return goog.net.XmlHttp();
+  return this.xmlHttpFactory_ ? this.xmlHttpFactory_.createInstance() :
+                                goog.net.XmlHttp();
```
mashedcode: See [comment below](https://github.com/google/closure-library/pull/797/files#r102348022). By the way this seems to be the only case where `goog.net.XmlHttp` is called in this module.`

closure/goog/net/xhrio.js
code-4-fun: I am using clang-format version 4.0.0 and also confirmed that I use `style=Google` flag while formatting. When I execute `clang-format -i -style=Google foo.js bar.js` for the first time before commit, it did reformat the complete file. I think, I should have used the `git diff***` variant of the command instead.

closure/goog/testing/net/xhrio_test.js
code-4-fun: yeah, I also realised, creating an `XmlHttp` instance in this mock is quite un-necessary. So, now we simply return and empty object. I think that makes sense.

closure/goog/net/xhrio.js
mashedcode: @code-4-fun Ohh right. The closure-library seems to still use [3.7.1](https://github.com/google/closure-library/blob/8b3e2795d3b2d40d67eacc2bfa02c30718eda04c/scripts/ci/install_closure_deps.sh#L6) at the moment.
@joeltine can you bump up the clang format version?
Note: The change was included in the clang release 3.8.0 but I do have trouble finding the commit that introduced this change in clang-format. (Hopefully this change was intentional.)

shicks:
I have to agree with @mashedcode here.  On the one hand, I agree that the functionality is somewhat useful.  On the other, each additional optional parameter comes at a pretty high API cost, and I'm not convinced the cost is worth the benefit given the other options already available.  If we had opted for using an object to "name" the optional parameters when this was originally implemented, then the API cost would be near zero and I'd have no problem merging this.  But as it stands now, I'd prefer to let this one go.

Additionally, I'd encourage you to look at modern browsers' built-in "fetch" API, which we hope will eventually supersede goog.net.XhrIo, since it offers a significantly cleaner interface, if it's supported in the browsers you need to support at least.

PR #801, opened 2017-02-27
Remove Sha1 from Readme

MarcusSorealheis:
Here it is harmless, but it could go far. Since SHA-2 is in the closure library, the README should default to using SHA-2 to prevent instilling bad habits that could have consequences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ require("google-closure-library");
 
 goog.require("goog.crypt.Sha1");
 
-var sha1 = new goog.crypt.Sha1();
-sha1.update("foobar");
-var hash = sha1.digest();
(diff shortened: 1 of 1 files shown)


MarcusSorealheis:
I signed it!

MarcusSorealheis:
run tests

mashedcode:
You need to squash your commits.
I don't think it makes a difference.
SHA-1 is fine as a hash function (git still uses it). It's just not considered secure.
By the way closure library SHA-2 implementation has more dependencies than the SHA-1 implementation.

MarcusSorealheis:
@mashedcode I will squash them. SHA-1 is only useful as a content identifier, which is its purpose on git. I think we should not use it for any other purpose.

PR #807, opened 2017-03-15
Fix push_latest_docs to export GIT_DIR correctly

shicks:
Previously we were getting errors about "src refspec gh-pages does not match any." because we were trying to push from the wrong repo.

RELNOTES=n/a


concavelenz:
LGTM

PR #810, opened 2017-03-19
Fix return type of IndexedDB.prototype.version

vobruba-martin:
Current browsers return a number. This was specced in [W3C Working Draft 06 December 2011](https://www.w3.org/TR/2011/WD-IndexedDB-20111206/#idl-def-IDBDatabase).
I don't know about any browser which return a string here.

This PR is prerequisite for google/closure-compiler#2035.

See
https://www.w3.org/TR/IndexedDB/#idl-def-IDBDatabase
https://github.com/google/closure-compiler/pull/2035#issuecomment-286847194

/cc @brad4d

### DIFF
--- a/closure/goog/db/indexeddb.js
+++ b/closure/goog/db/indexeddb.js
@@ -138,10 +138,11 @@ goog.db.IndexedDb.prototype.getName = function() {
 
 
 /**
- * @return {string} The current database version.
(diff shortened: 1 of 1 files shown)


vobruba-martin:
Forced Travis rebuild

shicks:
Testing internally to see if anything breaks.

PR #812, opened 2017-03-24
Disable IE testing

joeltine:
Temporarily disable IE8 until we can fix this "Modal dialog present" issue https://travis-ci.org/google/closure-library/builds/213615113

I've been working with Sauce Labs... but it hasn't been very helpful.


shicks:
Why 8 and 10, but not 9?

PR #816, opened 2017-04-01
allow en_ISO LOCALE to set DateTimeSymbols

### LINKED ISSUE #813: Can't use en_ISO DateTimeSymbols by setting goog.LOCALE
I'm not sure if this is intentional or not but there's no clause to use the `en_ISO` `DateTimeSymbols` if the LOCALE is set to `en_ISO`. Is there a reason for this or is it maybe just an oversight? (see [datetimesymbols.js](https://github.com/google/closure-library/blob/f68f494fd33b4ae7efa8f66701313eb56692fb80/closure/goog/i18n/datetimesymbols.js))

martinklepsch:
Not sure if it should be placed somewhere else because ISO is not a "language"... via #813


fabalbon:
Thanks for your pull request. #813 was fixed internally and will be available shortly.

Issue #821, opened 2017-04-20
testrunner.js uses deprecated goog.testing.TestCase.getTestResults


keghani:
Greetings from https://github.com/googlei18n/libphonenumber!

I'm getting errors compiling our JS library with the latest closure dependencies (closure-library, closure-compiler) because [testrunner.js](https://github.com/google/closure-library/blob/0592a220b21b53db3cb77caaaea69cadec20789d/closure/goog/testing/testrunner.js#L440) is making a call to deprecated [goog.testing.TestCase.getTestResults](https://github.com/google/closure-library/commit/acf652655b8f67ed36e01cdfff03765f1595c901#diff-f2db8d43347663cf7438b30cc1701fb0R712) which is deprecated; and we treat deprecation issues as errors ([our compilation flags](https://github.com/googlei18n/libphonenumber/blob/master/javascript/build.xml#L41)).

Is there a plan to update the code in testrunner.js?

The deprecation message is also in this repo's [Travis output](https://travis-ci.org/google/closure-library/builds/221324571#L3713), but probably not stopping builds because of [this compilation flag](https://github.com/google/closure-library/blob/e5a94170fbdc5eb759cfc4edfff4b55d95eb4fee/scripts/ci/compile_closure.sh#L15).

To reproduce:
```
git clone https://github.com/googlei18n/libphonenumber.git
cd libphonenumber
ln -s $CLOSURE_LIBRARY_DIR_PATH ../closure-library
ln -s $CLOSURE_COMPILER_DIR_PATH ../closure-compiler
ant -f javascript/build.xml compile
```

Output contains:
```
     [exec] /tmp/closure-library/closure/goog/testing/testrunner.js:440: ERROR - Property getTestResults of type goog.testing.TestCase has been deprecated.
     [exec]     return this.testCase.getTestResults();
     [exec]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     [exec] 
     [exec] 1 error(s), 0 warning(s), 97.6% typed
     [exec] calcdeps.py: JavaScript compilation failed.
```

joeltine:
No concrete plans that I know of. We're kind of understaffed atm and
cleaning up deprecated methods is rather low priority. Feel free to submit
a PR (or CL) if you'd like.

keghani:
Thanks @joeltine! I just realized the deprecated method is part of this project (d'oh) so sent a PR.

shicks:
@dankurka Can you take a look at this?

[...]

PR #822, opened 2017-04-28
Remove deprecation of method without alternative that's still in use in project

keghani:
Namely: https://github.com/google/closure-library/blob/0592a220b21b53db3cb77caaaea69cadec20789d/closure/goog/testing/testrunner.js#L440 is using this.

https://github.com/google/closure-library/issues/821


shicks:
@dankurka deprecated this - I'm not sure we want to simply undeprecate it.  Rather, we should be able to remove the deprecated usage from within our own library.  I believe we had previously tried to do this.

lararennie:
While that is true, he deprecated it *without giving an alternative* so it's not very simple to do that. In the meantime, this is breaking compilation, so as a short-term fix, rolling back the deprecation seems reasonable?

PR #830, opened 2017-06-14
Pull Closure Compiler from a known-green snapshot.

shicks:
Tests should pull from a known-green snapshot rather than building the latest version.

RELNOTES: n/a


shicks:
Not sure if Travis will run the new script - presumably it will: https://travis-ci.org/google/closure-library/builds/242668533?utm_source=github_status&utm_medium=notification

scripts/ci/install_closure_deps.sh
```diff
@@ -19,13 +20,16 @@ mv $CLANG_BUILD clang
 rm -f $CLANG_TAR
 
 # Install closure compiler and linter.
-if [ ! -d "closure-compiler" ]; then
-  git clone --depth 1 https://github.com/google/closure-compiler.git
+if [ ! -e "closure-compiler-1.0-SNAPSHOT.jar" ]; then
```
MatrixFrog: I think Travis effectively runs on a fresh machine each time, so I would expect this to always return true. If I'm wrong you just might compile using an older compiler, which should generally be fine I suppose.

MatrixFrog:
lgtm

PR #835, opened 2017-06-28
833

### LINKED ISSUE #833: goog.dom.forms getValue and setValue both require `type` to be set but not all HTML elements with value have a type
For example, `meter` and `progress` HTML elements have a `value` accessible via. `element.value` but not a `type`.

The default fallback for `getValue` and `setValue` should work for these elements.

Is there a reason to return at the start of these functions if elements don't have a type, because it seems unrelated to the whether the element might have a value?

thedavidmeister:
fixes #833

### DIFF
--- a/closure/goog/dom/forms.js
+++ b/closure/goog/dom/forms.js
@@ -320,16 +320,14 @@ goog.dom.forms.hasValueByName = function(form, name) {
 goog.dom.forms.getValue = function(el) {
   // Elements with a type may need more specialized logic.
   var type = /** @type {!HTMLInputElement} */ (el).type;
-  if (goog.isDef(type)) {
-    switch (type.toLowerCase()) {
-      case goog.dom.InputType.CHECKBOX:
-      case goog.dom.InputType.RADIO:
-        return goog.dom.forms.getInputChecked_(el);
-      case goog.dom.InputType.SELECT_ONE:
-        return goog.dom.forms.getSelectSingle_(el);
-      case goog.dom.InputType.SELECT_MULTIPLE:
-        return goog.dom.forms.getSelectMultiple_(el);
-    }
+  switch (goog.isDef(type) && type.toLowerCase()) {
+    case goog.dom.InputType.CHECKBOX:
+    case goog.dom.InputType.RADIO:
+      return goog.dom.forms.getInputChecked_(el);
+    case goog.dom.InputType.SELECT_ONE:
+      return goog.dom.forms.getSelectSingle_(el);
+    case goog.dom.InputType.SELECT_MULTIPLE:
+      return goog.dom.forms.getSelectMultiple_(el);
   }
   // Not every element with a value has a type, for example meter and progress.
   return goog.isDef(el.value) ? el.value : null;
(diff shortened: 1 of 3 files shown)


thedavidmeister:
I signed the CLA

closure/goog/dom/forms_test.js
```diff
@@ -417,6 +417,26 @@ function testSetValueInput() {
   assertEquals('', goog.dom.forms.getValue(el));
 }
 
+function testGetValueMeter() {
+  var min = 0;
+  var max = 3;
+  var val = 2.3;
+  var el = goog.dom.createDom('meter', {'min': min, 'max': max, 'value': val});
```
shicks: Let's inline these values rather than pulling them out as variables.  It should make them a lot easier to read.

closure/goog/dom/forms_test.js
```diff
@@ -417,6 +417,26 @@ function testSetValueInput() {
   assertEquals('', goog.dom.forms.getValue(el));
 }
 
+function testGetValueMeter() {
```
shicks: Does this work correctly in IE8?  My understanding is that IE8 doesn't natively support meter or progress elements, so I was expecting this to bail out early, say if el was undefined, or if its tagname != 'meter'.  Is that not an issue here?

closure/goog/dom/forms_test.js
```diff
@@ -417,6 +417,26 @@ function testSetValueInput() {
...
+}
+
+function testSetValueMeter() {
+  var min = 1;
+  var max = 5;
+  var el = goog.dom.createDom('meter', {'min': min, 'max': max, 'value': 3});
+
+  assertEquals(3, goog.dom.forms.getValue(el));
+
+  var newValue = +(Math.random() * (max - min) + min).toPrecision(3);
```
shicks: Avoid using random in tests - let's just pick a number and stick with that.

closure/goog/dom/forms.js
```diff
@@ -318,21 +318,21 @@ goog.dom.forms.hasValueByName = function(form, name) {
...
-    case goog.dom.InputType.CHECKBOX:
-    case goog.dom.InputType.RADIO:
-      return goog.dom.forms.getInputChecked_(el);
-    case goog.dom.InputType.SELECT_ONE:
-      return goog.dom.forms.getSelectSingle_(el);
-    case goog.dom.InputType.SELECT_MULTIPLE:
-      return goog.dom.forms.getSelectMultiple_(el);
-    default:
-      return goog.isDef(el.value) ? el.value : null;
+  if (goog.isDef(type)) {
```
shicks: Rather than indenting this an extra step we can simply fold the check in directly:

  switch (type && type.toLowerCase()) {

This will also allow to continue using the default branch.

closure/goog/dom/forms_test.js
thedavidmeister: why would you avoid using random in tests? if anything min and max should be randomised too.

closure/goog/dom/forms_test.js
thedavidmeister: maybe it is an issue, i can fix that up

closure/goog/dom/forms_test.js
thedavidmeister: ok

closure/goog/dom/forms.js
thedavidmeister: i can change this.

the double indenting is just how this was already done in `setValue`.

closure/goog/dom/forms_test.js
thedavidmeister: the tests don't seem to be running in IE8 on Travis from what i can tell in the logs.

https://travis-ci.org/google/closure-library/builds/247947020?utm_source=github_status&utm_medium=notification

I'm seeing chrome, ff, ie9 and ie11

I don't have a copy of ie8 handy, and tbh i'm not even sure how to run the tests for closure locally.

I'm not sure how to reproduce the failure somewhere that i can see, and then fix it. Is this something you could help with @shicks ?

shicks: I can try pulling it internally and seeing if our internal IE8 tests fail.

thedavidmeister: @shicks ok i think i've done as much as i can for this PR, let me know how you go

shicks: This seems to pass on IE8, so I'll try to push it through.

thedavidmeister: @shicks yay :)

PR #836, opened 2017-06-29
Fix lint_pull_request.sh to use linter pulled from snapshot.

shicks:
Ever since 84b8392 linting PRs has failed, since we no longer download and build the Closure Compiler repo from source.

### DIFF
--- a/scripts/ci/install_closure_deps.sh
+++ b/scripts/ci/install_closure_deps.sh
@@ -7,11 +7,25 @@ declare -r CLANG_VERSION="3.7.1"
 declare -r CLANG_BUILD="clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-14.04"
 declare -r CLANG_TAR="${CLANG_BUILD}.tar.xz"
 declare -r CLANG_URL="http://llvm.org/releases/${CLANG_VERSION}/${CLANG_TAR}"
-declare -r CLOSURE_COMPILER_REPO="https://oss.sonatype.org/content/repositories/snapshots/com/google/javascript/closure-compiler/1.0-SNAPSHOT/"
 
 set -ex
 
 cd ..
+
+# Fetches Closure Compiler components from oss.sonatype.org
+fetch () {
+  local name="$1"
+  local snapshots="https://oss.sonatype.org/content/repositories/snapshots"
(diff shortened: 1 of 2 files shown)


scripts/ci/install_closure_deps.sh
```diff
@@ -7,25 +7,37 @@ declare -r CLANG_VERSION="3.7.1"
...
-declare -r CLOSURE_COMPILER_REPO="https://oss.sonatype.org/content/repositories/snapshots/com/google/javascript/closure-compiler/1.0-SNAPSHOT/"
 
 set -ex
 
 cd ..
 
+# Fetches Closure Compiler components from oss.sonatype.org
+fetch () {
+  local name="${1%-1.0-SNAPSHOT.jar}"
+  [ "$name" != "$1" ]
```
MatrixFrog: I don't understand what this line is doing

scripts/ci/install_closure_deps.sh
```diff
@@ -7,25 +7,37 @@ declare -r CLANG_VERSION="3.7.1"
...
+  local snapshots="https://oss.sonatype.org/content/repositories/snapshots"
+  local repo="$snapshots/com/google/javascript"
+  local dir="$repo/$name/1.0-SNAPSHOT"
+  : "closure-compiler/1.0-SNAPSHOT/"
```
MatrixFrog: or this one

scripts/ci/install_closure_deps.sh
shicks: It's an assertion.  If `$name == $1` then that means that the argument didn't have the expected suffix, so something has gone wrong.

scripts/ci/install_closure_deps.sh
shicks: This does absolutely nothing.  I have no idea why it's in there.  Aside from a vague sense that I was commenting something out and must have accidentally hit `:` instead of `#`...

scripts/ci/install_closure_deps.sh
shicks: It's an assertion.  If they're equal then the suffix-removal above didn't work.

That said, I could just as well leave off the `-1.0-SNAPSHOT.jar` from the argument and then not need to worry about that possibility.

MatrixFrog:
lgtm

PR #838, opened 2017-07-01
Update package.json to latest NPM (20170626.0.0)

### LINKED ISSUE #837: package.json version incorrect and out-of-date
The version specifier in the package.json isn't valid semver - need to leave off the leading 'v' character.

Also, it's significantly out of date with the version published to npm. Is this not getting MOE sync'd?

shicks:
Fixes #837

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "google-closure-library",
   "description": "Google's common JavaScript library",
-  "version": "v20170409.0.0",
+  "version": "20170626.0.0",
   "repository": {
     "type": "git",
     "url": "https://github.com/google/closure-library.git"


shicks:
@ChadKillingsworth Is this what it should be?

ChadKillingsworth:
Yes exactly! For reference, here's the compiler version: https://github.com/google/closure-compiler-npm/blob/master/package.json#L3

I'm not sure how the npm publication works internal to Google either. For the compiler, we have travis auto publish it to npm on every successful build. Since you can't publish the same version twice, travis only publishes new versions when the package.json changes.

shicks:
Oh, that would be awesome to get that set up.  Currently we have a script that @joeltine wrote to manually push to NPM.  I believe the script is also supposed to (separately) create a pull request to update package.json, but it's been erroring at that step for the last several months, and nobody's taken the time to figure out why or how to fix it.

PR #839, opened 2017-07-01
Add more specific annotation for goog.array.bucket

DavidANeil:


shicks:
A few formatting issues, but we can take care of that on import.

I'm a bit worried that we'll expose some bad typing on at least one or two internal projects with this change, since introducing better typing is always dangerous.  But I'll run it through our internal queue and see what happens.

shicks:
This did expose a handful of type errors.  Once the fixes have been submitted, I will pull this change.

shicks:
We did have some breakages, but they've been fixed and the change is now being merged internally.

shicks:
This is now being merged.

Issue #808, opened 2017-03-16
dom/TagName.MAIN missing?


binarykitchen:
I think the tag name `MAIN` is missing in the enums.

mashedcode:
Great that you look at the source code and make comments.
But it would be better if you'd figure out what commit actually introduced HTML5 tags and from there on you could go figure out why it's missing.
You could then open a PR and fix this stating why this was missing and why you need it and also add all other tag names that are missing while you're at it.

For example the HTML5 tag names where added to this project nearly five years ago in 03451dec299109612bf33376809ef4edca2a6883 and they were taken from the [HTML5 draft specification](https://web.archive.org/web/20120704221326/http://dev.w3.org/html5/spec/section-index.html) that is linked at [tagname.js#L22](
https://github.com/google/closure-library/blob/03451dec299109612bf33376809ef4edca2a6883/closure/goog/dom/tagname.js#L22). In that draft the `main` tag was not proposed yet therefore it was not added back then.

You're welcome to open a PR and add tag names according to the current specification.

I would be very astonished if one of the guys who maintain this project as part of their job would look into it in that much detail based on a single sentence you drop in an issue. So please try to do some research on your own.

mashedcode:
@b-strauss @binarykitchen and everyone else sorry for my toxic behavior. @pauldraper thanks for taking care of this.

PR #853, opened 2017-10-31
Add missing HTML5.1 elements to dom/tagname.js.

PaulCapron:
These are `<main>`, `<menuitem>`, `<picture>`, and `<rtc>`.

They were absent in the HTML5 draft on which `dom/tagname.js` was based.

As of today, HTML5.1 2nd edition is the current W3C recommendation, linked from https://www.w3.org/TR/html/.

This commit closes GitHub issue #808.


shicks:
Can you update this array to include any of the new tagnames that default to display: block, please?

https://github.com/google/closure-library/blob/master/closure/goog/editor/node_test.js#L494

shicks:
Wrong location - looks like it needed to be added to the list in `node.js`.  I've added it to the internal PR.

PR #854, opened 2017-11-22
add goog.date.DateTime.timeEquals function

bambusekd:
I came to a problem, where I needed to compare two DateTimes, but I was only interested in comparing the time part regardless of the year/mont/date value. Since I did not found any existing function that could do that, I have added a **timeEquals** method to **goog.date.DateTime**.

Do you think such a function might be of a good use for someone else or is it just me needing it? Or is there a better way to check whether two times are equal?

Thanks for any comments and ideas.

### DIFF
--- a/closure/goog/date/date.js
+++ b/closure/goog/date/date.js
@@ -1665,6 +1665,18 @@ goog.date.DateTime.prototype.equals = function(other) {
 
 
 /**
+ * Tests whether given time part of datetime is exactly equal to time part of this DateTime.
+ *
+ * @param {goog.date.DateTime} other The datetime to compare.
+ * @return {boolean} Whether the given time part of datetime is exactly equal to time part of this one.
+ * @override
+ */
+goog.date.DateTime.prototype.timeEquals = function(other) {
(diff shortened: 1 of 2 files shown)


closure/goog/date/date.js
```diff
@@ -1664,6 +1664,18 @@ goog.date.DateTime.prototype.equals = function(other) {
...
 
 
+/**
+ * Tests whether given time part of datetime is exactly equal to time part of this DateTime.
+ *
+ * @param {goog.date.DateTime} other The datetime to compare.
+ * @return {boolean} Whether the given time part of datetime is exactly equal to time part of this one.
+ * @override
+ */
+goog.date.DateTime.prototype.timeEquals = function(other) {
```
jplaisted: My two cents:

- There's a bunch of isSame*blah* static methods. This fits in with those.
- Using "time" by itself is confusing here. Perhaps refer to it as time ignore date? e.g. isSameIgnoreDateMonthAndYear or something?

shicks:
Closing this after there's been no response from the author in nearly two years.  If you want to continue pursuing this, please respond to the review comments and reopen.

PR #866, opened 2018-01-08
Don't register mouseover listeners if control has HOVER disabled

ribrdb:


shicks:
This looks like just an optimization.  If the behavior is actually wrong, we can consider merging this (in which case please reopen it), but in general goog.ui is in maintenance mode and is not recommended for any current or future development, so we really don't want to be changing it if at all possible.

Issue #849, opened 2017-10-27
[Python3 ] closurebuilder.py: TypeError: a bytes-like object is required, not 'str'


Morabaraba:
Hi

I got a error using `closurebuilder.py` with python 3.5 and 3.6:

```sh
Traceback (most recent call last):
  File "closure-library/closure/bin/build/closurebuilder.py", line 293, in <module>
    main()
  File "closure-library/closure/bin/build/closurebuilder.py", line 282, in main
    compiler_flags=options.compiler_flags)
  File "/static/closure/closure-library/closure/bin/build/jscompiler.py", line 151, in Compile
    flags_file = _GetFlagFile(source_paths, compiler_flags)
  File "/static/closure/closure-library/closure/bin/build/jscompiler.py", line 122, in _GetFlagFile
    flags_file.write(' '.join(args))
  File "/venv/lib/python3.6/tempfile.py", line 622, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'
```

If I change the first line in `closurebuilder.py` to `#!/usr/bin/env python2` all work as expected.

I did see the migrate warning, so I don't think anybody want to spend time on this?

Maybe make the `closurebuilder.py` tool a python2 tool only and hardcode it to `python2`?

shicks:
Please feel free to send a PR that does this.

PR #868, opened 2018-01-10
Fix closurebuilder.py compilerflags on python >= 3.5

mbeijen:
Closes https://github.com/google/closure-library/issues/849

While I created this small fix, I also ran into a deprecation warning in python `assertEquals()` calls when running the tests on python 2.7, 3.4, 3.5, 3.6, so I also fixed those.


shicks:
Currently reviewing internally as cr/272555293

brad4d:
Internal Google issue http://b/142064794 created

mbeijen:
Applied in 03c3d88f21abbc8cff02f63912ef2b046a9be079

PR #879, opened 2018-02-20
Fix a type warning for textContent

koba04:
Currently, I see the following warning when compiling Closure Library with ADVANCED mode.
This PR is to fix this, which is casting `number` to `string` by concatenating an empty string.

```
WARNING in ./node_modules/google-closure-library/closure/goog/dom/dom.js:1759 from closure-compiler: The right side in the assignment is not a subtype of the left side.
Expected : string
Found    : (number|string)
More details:
The found type is a union that includes an unexpected type: number
    node.textContent = text;
    ^^^^^^^^^^^^^^^^^^^^^^^
```

Thanks!

### DIFF
--- a/closure/goog/dom/dom.js
+++ b/closure/goog/dom/dom.js
@@ -1756,7 +1756,7 @@ goog.dom.setTextContent = function(node, text) {
       'goog.dom.setTextContent expects a non-null value for node');
 
   if ('textContent' in node) {
-    node.textContent = text;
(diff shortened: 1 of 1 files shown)


koba04:
I signed it!

nreid260:
This should no longer be an issue now that the extern for `Node::textContent` is annotated `@implictCast`. Thanks for the effort though!

PR #885, opened 2018-03-14
wheelhandler demo: fix `WHEEL` and `DeltaMode` references

julen:
These are bogus references which prevent the demo from working.


julen:
I signed it!

PR #919, opened 2018-08-15
Fix the path to match the key in depsByPath Map

luzhang:
In the depsByPath Map,  the key uses the path defined in parser.parseFile().
`this.moduleResolver.resolve(i.from.path, i.symOrPath)` returns an absolute path, which causes NOT FOUND error after calling this.depsByPath.get(). The issue will be resolved by replacing it with `i.from.path`.

E.g. 
depsByPath:  Map {
  'src/js/second.js' => Dependency {
  type: 'es6 module',
  path: 'src/js/second.js',
  closureSymbols: [],
  imports: [],
  language: 'es6' }}

### DIFF
--- a/closure-deps/lib/depgraph.js
+++ b/closure-deps/lib/depgraph.js
@@ -333,8 +333,7 @@ class Graph {
   resolve_(i) {
     return i instanceof GoogRequire ?
         this.depsBySymbol.get(i.symOrPath) :
-        this.depsByPath.get(
-            this.moduleResolver.resolve(i.from.path, i.symOrPath));
+        this.depsByPath.get(i.from.path);
   }
 
   /**
--- a/closure-deps/lib/depgraph.js
+++ b/closure-deps/lib/depgraph.js
@@ -193,7 +193,7 @@ class ModuleResolver {
 class PathModuleResolver {
   /** @override */
   resolve(fromPath, importSpec) {
-    return path.join(path.dirname(fromPath), importSpec);
+    return path.resolve(path.dirname(fromPath), importSpec);
   }
 }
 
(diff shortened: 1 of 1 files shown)


jplaisted:
This will just disable custom resolution, which isn't right (notice after this change that the module resolve ris unused). It seems like the correct thing is to have the keys of the map be absolute paths.

jplaisted:
Alternatively use `path.join` rather than `path.resolve` in the `PathModuleResolver`.

closure-deps/lib/depgraph.js
```diff
@@ -231,7 +231,8 @@ class Graph {
       if (this.depsByPath.has(dep.path)) {
         throw new Error('File registered twice? ' + dep.path);
       }
-      this.depsByPath.set(dep.path, dep);
+      this.depsByPath.set(
+          path.join(path.dirname(dep.path), path.basename(dep.path)), dep);
```
jplaisted: This doesn't make the key an absolute path. This code should just return end up being `dep.path` again. `path.dirname('./a/b')` will be `./a`, fyi.

`path.resolve` will make a path absolute. `path.join` won't.

`path.join('../a', 'b')` will be `../a/b`. `path.resolve('../a', 'b')` will be some absolute path ending with `/a/b`.

https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths
https://nodejs.org/docs/latest/api/path.html#path_path_join_paths


tl;dr I think that when a `Dependency` is created you should say `this.path = path.resolve(filePath)` in the constructor. I think that should force all `Dependency`s to have absolute paths, meaning this map will use absolute paths, and `PathModuleResolver` can remain unchanged.

luzhang:
Sorry for spamming the commits, I barely use the github UI directly to make commits, seems it sent out 3 commits instead of 1.

PR #921, opened 2018-08-16
Added two currencies that were missing

TimovanderKamp:


nreid260:
Will be partially included in the next release. We decided against adding ZWD.

PR #929, opened 2018-09-20
Add gulp plugin for Closure deps writer

luzhang:


closure-deps/lib/gulp/index.js
```diff
@@ -0,0 +1,51 @@
+const Buffer = require('buffer').Buffer;
+const {parser, depGraph, depFile} = require('google-closure-deps');
+const path = require('path');
+const PluginError = require('plugin-error');
+const through = require('through2');
```
jplaisted: Some of these (not sure which) are not native to Node. You need to update package.json for these new dependencies.

But maybe this would work better as a separate Node package to keep this package's dependencies minimal?

luzhang: Thanks for catching it, I have added those dependencies into package.json.
I feel using a single package.json for the module is easier to maintain and use, especially when there are only 3 dependencies, probably not necessary to create a separate package.json.

jplaisted:
We've discussed elsewhere that this works better as its own package (that can have a dependency on google-closure-deps).

PR #930, opened 2018-09-20
Efficiently copying AVL trees.

kevin-dp:
Currently the `AvlTree` does not provide a way to efficiently make a deep copy of the tree.
Hence, programmers need to fetch all values, manually deep copy the values, and then insert them in the new tree:

```
const tree = new goog.structs.AvlTree();
// ...
const copy = new goog.structs.AvlTree();

// Inneficient deep copy: O(n * k * log(n)),
// where k is the time complexity to copy a value
tree.getValues()
    .forEach(v => copy.add(JSON.parse(JSON.stringify(v))));
```

The above approach is inneficient since it is in `O(n * k * log(n))`, where `k` is the time complexity to copy a value contained by the tree.

This pull request extends the `AvlTree` with a `copy` method which deep copies the tree in `O(n * k)`.

### DIFF
--- a/closure/goog/structs/avltree.js
+++ b/closure/goog/structs/avltree.js
@@ -900,8 +900,9 @@ Node.prototype.fixHeight = function() {
 };
 
 /**
- * @callback copyCb
- * @param {T}
+ * @name copyCb
+ * @function
+ * @param {T} value - A value contained by the tree.
  * @return {T}
  */
 
--- a/closure/goog/structs/avltree.js
+++ b/closure/goog/structs/avltree.js
@@ -929,13 +929,18 @@ AvlTree.prototype.copy = function(copy) {
 };
 
 /**
+ * @typedef SubTree
+ * @type {object}
+ * @property {Node<T>} root - Root node of the subtree.
+ * @property {Node<T>} leftMost - Node with the smallest value in the subtree.
+ * @property {Node<T>} rightMost - Node with the largest value in the subtree.
+ */
+
+/**
  * Copies a node.
  * @param {Node<T>} parent - The parent of this node.
- * @param {copyCb=} copy - Function used to copy the values contained by the tree.
- * @return {Object} treeInfo - Information about the subtree.
- * @return {Node<T>} treeInfo.copy - Copy of the root node of this subtree.
- * @return {Node<T>} treeInfo.leftMost - Copy of the node with the smallest value in this subtree.
- * @return {Node<T>} treeInfo.rightMost - Copy of the node with the largest value in this subtree.
+ * @param {copyCb=} copy - Function used to copy the values contained by the subtree.
+ * @return {SubTree} subtree - Information about the copied subtree.
  */
 Node.prototype.copy = function(parent, copy) {
   const val = typeof copy === "function" ? copy(this.value) : JSON.parse(JSON.stringify(this.value));
--- a/closure/goog/structs/avltree.js
+++ b/closure/goog/structs/avltree.js
@@ -900,15 +900,8 @@ Node.prototype.fixHeight = function() {
 };
 
 /**
- * @name copyCb
- * @function
- * @param {T} value - A value contained by the tree.
- * @return {T}
- */
-
-/**
  * Copies the AVL tree.
- * @param {copyCb=} copy - Function used to copy the values contained by the tree.
+ * @param {function(T):T} [copy] - Function used to copy the values contained by the tree.
  * @return {AvlTree<T>}
  */
 AvlTree.prototype.copy = function(copy) {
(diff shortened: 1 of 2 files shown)


closure/goog/structs/avltree.js
```diff
@@ -898,4 +899,47 @@ Node.prototype.fixHeight = function() {
...
+  tree.root_       = copy;
+  tree.comparator_ = this.comparator_;
+  tree.minNode_    = leftMost;
+  tree.maxNode_    = rightMost;
+
+  return tree;
+};
+
+Node.prototype.copy = function(parent) {
+  const val  = JSON.parse(JSON.stringify(this.value)); // deep copy
```
jplaisted: This only works for basic objects (doesn't copy methods) and will wipe out the prototype chain.

```
class A {
  foo() { console.log('foo'); }
}
const copy = JSON.parse(JSON.stringify(new A()));
copy instanceof A; // false
copy.foo(); // error
```

Maybe have `copy` take an optional copy function? It can use JSON as the default.

```
class A {
  copy() {
    return new A();
  }
}
tree.copy((a) => a.copy());
```

closure/goog/structs/avltree.js
```diff
@@ -898,4 +899,47 @@ Node.prototype.fixHeight = function() {
...
+ * Returns a deep copy of the tree in O(n * k),
+ * where k is the time complexity to copy a node.
+ * @return {AvlTree}
+ */
+AvlTree.prototype.copy = function() {
+  const tree = new AvlTree();
+
+  // Copy instance properties
+  const {copy, leftMost, rightMost} = this.root_.copy();
+  tree.root_       = copy;
```
jplaisted: Single space here and below.

closure/goog/structs/avltree.js
```diff
@@ -898,4 +899,47 @@ Node.prototype.fixHeight = function() {
...
+
+Node.prototype.copy = function(parent) {
+  const val  = JSON.parse(JSON.stringify(this.value)); // deep copy
+  const node = new Node(val, parent);
+
+  // Copy all properties
+  node.count  = this.count;
+  node.height = this.height;
+
+  var minNode = node, maxNode = node;
```
jplaisted: If you want to use ES6 in this new code (which I'm not sure will land - but I'll fix it when I pull this in if it can't) then please be consistent and use `const` or `let` here..

closure/goog/structs/avltree.js
```diff
@@ -898,4 +899,47 @@ Node.prototype.fixHeight = function() {
...
+
+  if (this.left) {
+      const {copy, leftMost, rightMost} = this.left.copy(node);
```
jplaisted: Google style is indent 2 spaces.

closure/goog/structs/avltree.js
```diff
@@ -898,4 +899,47 @@ Node.prototype.fixHeight = function() {
...
+/**
+ * Returns a deep copy of the tree in O(n * k),
+ * where k is the time complexity to copy a node.
+ * @return {AvlTree}
+ */
+AvlTree.prototype.copy = function() {
+  const tree = new AvlTree();
+
+  // Copy instance properties
+  const {copy, leftMost, rightMost} = this.root_.copy();
```
jplaisted: `this.root_` is `null` for empty trees (add a unit tests for this too please).

closure/goog/structs/avltree_test.js
```diff
@@ -630,6 +630,45 @@ function testRemoveRightLeftCase() {
...
+
+  tree = tree.copy();
+
+  //   100                            100                             112
+  //   / \                            / \                            /  \
+  // 50   150   Right Rotate (150)   50  112      Left Rotate(100)  100  150
+  //     /   \   - - - - - - - ->           \     - - - - - - - ->  /   /   \
+  //    112  200                            150                    50   125  200
+  //     \                                  /  \
+  //     125                               125 200
```
jplaisted: Most of the above is copy and pasted from another unit test and isn't really relevant. At the very least delete this comment, we aren't really testing the rotation here.

closure/goog/structs/avltree.js
kevin-dp: By default, `JSON.stringify` will copy the value using the value's `toJSON` method.
Hence, no need to provide an optional copy function.
However, the problem lies in the deserialization.
As u pointed out this way of deep copying the object wipes out the prototype chain.
Hence, i suggest to either take an optional parsing function, or take both an optional copy function and an optional parsing function. What do u prefer?

kevin-dp: Indeed!
I will add this.

kevin-dp: I introduced an error in my latest commit.
I don't have the time now but i will fix this ASAP.

jplaisted:
I'm going to merge this with a few edits. The only one that impacts functionality is that it won't use `JSON` by default at all. The default behavior (no copy function) will be assignment copy (e.g. the `=` operator, so it won't be a deep copy by default). You can copy with `JSON` if you like.

PR #862, opened 2018-01-01
Add .npmignore and reduce npm package size

teppeis:
`npm pack`ed tgz is reduced from 4.9MB to 3.1MB.
The diff list is https://gist.github.com/teppeis/743264bf135ef712daaae9872a93d091

PR #934, opened 2018-10-24
Fix google-closure-deps

teppeis:
- Update `google-closure-compiler` to pass tests using `goog.declareModuleId()` (since v20181008)
- Downgrade `google-closure-library` in devDeps to pass `closuremakedeps_test.js`
    - #862 removes useless files like `*_test.js` from published npm package in v20180805
- Add missing `jasmine-diff` to devDeps
- Run tests in Travis CI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 
 node_modules/
 closure-deps/node_modules/
+package-lock.json
--- a/closure-deps/package.json
+++ b/closure-deps/package.json
@@ -33,11 +33,12 @@
   },
   "homepage": "https://developers.google.com/closure/library/",
   "devDependencies": {
-    "google-closure-library": "^20180805.0.0",
-    "jasmine": "^2.9.0"
+    "google-closure-library": "^20180716.0.0",
+    "jasmine": "^2.9.0",
+    "jasmine-diff": "^0.1.3"
   },
   "dependencies": {
     "argparse": "^1.0.9",
-    "google-closure-compiler": "^20180805.0.0"
+    "google-closure-compiler": "^20181008.0.0"
   }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@
 # See the LICENSE file for details.
 
 language: node_js
+node_js:
+  - "8"
 sudo: required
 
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@
 
 node_modules/
 closure-deps/node_modules/
+package-lock.json
```
teppeis: It seems that lock files were not intentionally added, so I ignored.

closure-deps/package.json
```diff
@@ -33,11 +33,12 @@
...
-    "google-closure-library": "^20180805.0.0",
-    "jasmine": "^2.9.0"
+    "google-closure-library": "^20180716.0.0",
+    "jasmine": "^2.9.0",
+    "jasmine-diff": "^0.1.3"
   },
   "dependencies": {
     "argparse": "^1.0.9",
-    "google-closure-compiler": "^20180805.0.0"
+    "google-closure-compiler": "^20181008.0.0"
```
teppeis: Is this better?

```suggestion
    "google-closure-compiler": ">=20181008.0.0"
```

.gitignore
jplaisted: I'll think about adding this in the future, for this and the library.

closure-deps/package.json
jplaisted: SG

jplaisted:
Thanks for this! Adding the tests will help keep us in a better state in the future :+1:

closure-deps/package.json
```diff
@@ -33,11 +33,12 @@
   },
   "homepage": "https://developers.google.com/closure/library/",
   "devDependencies": {
-    "google-closure-library": "^20180805.0.0",
-    "jasmine": "^2.9.0"
+    "google-closure-library": "^20180716.0.0",
```
jplaisted: This is probably fine for now since it isn't too old. I'll make a PR in the future to have the closuremakedeps_test ignore *_test.js files in the expected output, and then we can use the head library again.

teppeis: fixed https://github.com/google/closure-library/commit/8e98199d401884257bc0758f3a35e173d0c178b5

PR #935, opened 2018-10-25
demo: fix break in dragdrop

kindy:
caused by `alert` when drag in list1.


nreid260:
What specifically does this fix? It looks like these alerts convey information as part of the demo, and would be fine in this context. Additionally, this change only eliminates one of many alerts.

shicks:
Closing as inactive.  Please reopen with more information if you wish to pursue this further.

PR #939, opened 2018-11-06
Add yargs as a dependency

jplaisted:
Planning on switching from argparse to yargs:

1) Yargs is much nicer to use
2) Yargs is more popular
3) Yargs has better documentation
4) Yargs has a cute pirate theme

### DIFF
--- a/closure-deps/package.json
+++ b/closure-deps/package.json
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "argparse": "^1.0.9",
-    "google-closure-compiler": "^20181008.0.0"
+    "google-closure-compiler": "^20181008.0.0",
+    "yargs": "^12.0.2"
   }
 }


teppeis:
@jplaisted yargs has been added to the deps, but it's not used yet.
Is switching working in progress? or can I work on it?

teppeis:
Also, yargs v12 is not latest now. v13 is the latest.

jplaisted:
Thanks for pinging. I needed to make this PR and then continue with the work, which I clearly left hanging (it would've ideal to do it in the same PR...). For now due to some technical restrictions we need to use v12.

teppeis:
@jplaisted Great.
If "some technical restrictions" means dropping support for Node v6, it's reverted recently in v13.2.2 https://github.com/yargs/yargs/pull/1308

PR #961, opened 2019-02-21
New NPM Release v20190215.0.0

blickly:


teppeis:
@blickly Please merge and publish new version to npm!

blickly:
Sorry I missed that! I'm trying to both do a library release for the first time and better document the steps so that we're more consistent with the closure library releases; it looks like I missed the `npm publism` step in my instructions.

I think it's released now (https://www.npmjs.com/package/google-closure-library) and that it's not necessary to merge the http://go/gh/google/closure-library/tree/npm-v20190215.0.0 branch into master.  Let me know if any of the assumptions I've made aren't true.

Thanks for catching!

teppeis:
@blickly Thanks!

teppeis:
@blickly closure-deps is not updated
https://www.npmjs.com/package/google-closure-deps

blickly:
@jplaisted, do you know what that package is and how it is released?

jplaisted:
https://www.npmjs.com/package/google-closure-library

I think we have a script to release it internally. But should be as simple as go to the commit the release was made at an run `npm publish`

blickly:
OK, I think this should be (finally) released.

PR #964, opened 2019-03-19
Fix closure-deps to allow latest compiler

teppeis:
Currently closure-deps allows only closure-compiler v20181008 and doesn't follow latest compiler.
Also if I use latest compiler and closure-deps in my project, the node_modules have two compiler versions.
This PR widdens the range of compiler version for the problem.

Another solution is updating the compiler version in the deps and publishing closure-deps for every compiler release.

### DIFF
--- a/closure-deps/package.json
+++ b/closure-deps/package.json
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "argparse": "^1.0.9",
-    "google-closure-compiler": "^20181008.0.0",
+    "google-closure-compiler": ">=20181008.0.0",
     "yargs": "^12.0.2"
   }
 }


teppeis:
The CI job failed in ./scripts/ci/generate_latest_docs.sh, but it's not related to this PR.

jplaisted:
Thanks! I'll take a look at making sure this stays in lock step with the library's dep. For now we can just accept this PR.

jplaisted:
This should be fixed with our next release, I don't think we need to merge this PR (why merge this if we won't do a release on it).

Sadly our release script is internal (I don't think it needs to be), so I can't link the fix :/

jplaisted:
But thanks for pointing this out!

teppeis:
Thanks!

PR #971, opened 2019-04-20
Typed array equals

### LINKED ISSUE #969: testSuite: assertObjectEquals and TypedArrays
Hi,

I'm trying to fix test cases from google/tink#190, and have found several failures around the testing of TypedArrays.

Basically, the test expects things like the following to pass:

```javascript
goog.module('tink.TypedArrayTest');
goog.setTestOnly('tink.TypedArrayTest');
const testSuite = goog.require('goog.testing.testSuite');
testSuite({
  testTypedArrays() {
    const buf1 = new ArrayBuffer(3);
    const buf2 = new ArrayBuffer(4);
    const ta1 = new Uint8Array(buf1, 0, 3);
    const ta2 = new Uint8Array(buf2, 1, 3);
    for (let i = 0; i < ta1.length; i ++) {
      ta1[i] = ta2[i] = i;
    }
    assertObjectEquals(ta1, ta2);
  }
});
```

but when I run it, I get:

```
-> 21:58:40.160 : Starting tests: Untitled Test Case
-> 21:58:40.160 : Running test: testTypedArrays
-> Expected <[object Uint8Array]> (Object) but was <[object Uint8Array]> (Object)
->    byteOffset: Expected <0> (Number) but was <1> (Number)
   buffer.byteLength: Expected <3> (Number) but was <4> (Number)
-> 21:58:40.164 : testRead : FAILED
```

I would argue that when testing TypedArray values, the user probably cares about value-equality (i.e. same array type, length, and contents), and not the attributes related to the backing buffer, much like the way Date objects are compared.

Does this warrant the introduction of a new assertion, say `assertContentEquals(ta1, ta2)`, or do you think it should be handled differently?  I can't do this easily in the project's code, as the TypedArrays being compared are nested deeply within other objects.  So,  I need to have something recursive that handles TypedArrays specially.

Please let me know what the design should be, and I will be willing to cook up a PR to do the right thing.

Thanks,
Michael.

michaelfig:
Fixes #969.

### DIFF
--- a/closure/goog/testing/asserts.js
+++ b/closure/goog/testing/asserts.js
@@ -47,6 +47,7 @@ var PredicateFunctionType;
 var PRIMITIVE_EQUALITY_PREDICATES = {
   'String': DOUBLE_EQUALITY_PREDICATE,
   'Number': DOUBLE_EQUALITY_PREDICATE,
+  'Bigint': DOUBLE_EQUALITY_PREDICATE,
   'Boolean': DOUBLE_EQUALITY_PREDICATE,
   'Date': function(date1, date2) { return date1.getTime() == date2.getTime(); },
   'RegExp': TO_STRING_EQUALITY_PREDICATE,
@@ -665,6 +666,24 @@ goog.testing.asserts.EQUALITY_PREDICATE_VARS_ARE_EQUAL = '';
 
 
 /**
+ * @const {!Object<string, boolean>}
+ */
+goog.testing.asserts.ARRAY_TYPES = {
+  'Array': true,
+  'Float32Array': true,
+  'Float64Array': true,
+  'Int8Array': true,
+  'Int16Array': true,
+  'Int32Array': true,
+  'Uint8Array': true,
+  'Uint8ClampedArray': true,
+  'Uint16Array': true,
+  'Uint32Array': true,
+  'BigInt64Array': true,
+  'BigUint64Array': true
+};
+
+/**
  * Determines if two items of any type match, and formulates an error message
  * if not.
  * @param {*} expected Expected argument to match.
@@ -749,7 +768,7 @@ goog.testing.asserts.findDifferences = function(
     var typeOfVar2 = _trueTypeOf(var2);
 
     if (typeOfVar1 == typeOfVar2) {
-      var isArray = typeOfVar1 == 'Array';
+      var isArray = goog.testing.asserts.ARRAY_TYPES[typeOfVar1];
       var errorMessage = equalityPredicate(typeOfVar1, var1, var2);
       if (errorMessage !=
           goog.testing.asserts.EQUALITY_PREDICATE_CANT_PROCESS) {
--- a/closure/goog/testing/asserts_test.js
+++ b/closure/goog/testing/asserts_test.js
@@ -357,6 +357,8 @@ function testAssertObjectNotEquals() {
 
   assertObjectNotEquals(new Map([['a', '1']]), new Map([['b', '1']]));
   assertObjectNotEquals(new Set(['a', 'b']), new Set(['a']));
+
+  assertObjectNotEquals(new Uint32Array([1, 2, 3]), new Uint32Array([1, 4, 3]));
 }
 
 function testAssertObjectEquals2() {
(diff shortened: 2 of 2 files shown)


michaelfig:
@shicks, could you please review?

closure/goog/testing/asserts.js
```diff
@@ -664,6 +665,24 @@ goog.testing.asserts.EQUALITY_PREDICATE_CANT_PROCESS = null;
 goog.testing.asserts.EQUALITY_PREDICATE_VARS_ARE_EQUAL = '';
 
 
+/**
+* @const {!Object<string, boolean>}
+ */
+var arrayTypes = goog.testing.asserts.ARRAY_TYPES = {
```
shicks: I don't think a top-level `var` is appropriate here.  This should only be the qualified name.

It's unfortunate that we have these unqualified names at all in this file, but we shouldn't be adding to them.

closure/goog/testing/asserts_test.js
```diff
@@ -436,9 +438,61 @@ function testAssertObjectEqualsMap() {
...
+    'Uint32Array',
+    'BigInt64Array',
+    'BigUint64Array',
+  ];
+
+  var prior, priorName;
+  for (var i = 0; i < typedArrays.length; ++i) {
+    var name = typedArrays[i];
+    var arrType = window[name];
+    if (arrType) {
```
shicks: (nit) if (!arrType) continue;

closure/goog/testing/asserts_test.js
```diff
@@ -436,9 +438,61 @@ function testAssertObjectEqualsMap() {
...
+      var arr2 = new arrType(buf2, arrType.BYTES_PER_ELEMENT, 3);
+
+      // Fill them with the same contents.
+      arr1[0] = arr2[0] = construct(1);
+      arr1[1] = arr2[1] = construct(2.5);
+      arr1[2] = arr2[2] = construct(-3);
+      assertObjectEquals(name + 's should be equal', arr1, arr2);
+
+      mutate(arr2);
+      assertThrows(name + 's should not be equal',
```
shicks: I think this needs to be assertThrowsJsUnitException.  I see that we don't have that elsewhere in this file, which leads me to believe we turn off check that it enables globally, but it still makes sense to do the right thing here.

closure/goog/testing/asserts_test.js
```diff
@@ -436,9 +438,61 @@ function testAssertObjectEqualsMap() {
...
-  assertThrows('sets should not be equal', function() {
-    assertObjectEquals(map1, map2);
-  });
+  assertThrows('maps should not be equal',
+               function() { assertObjectEquals(map1, map2); });
+}
+
+
+function testAssertObjectEqualsTypedArray() {
+  var typedArrays = [
```
shicks: Please use `let`, `const`, and arrow functions throughout this test case, as appropriate.

closure/goog/testing/asserts_test.js
```diff
@@ -436,9 +438,61 @@ function testAssertObjectEqualsMap() {
...
+    'Int32Array',
+    'Uint8Array',
+    'Uint8ClampedArray',
+    'Uint16Array',
+    'Uint32Array',
+    'BigInt64Array',
+    'BigUint64Array',
+  ];
+
+  var prior, priorName;
```
shicks: May be worth a comment that we're saving the prior in order to check that the same array of different types is not equal?

closure/goog/testing/asserts_test.js
```diff
@@ -436,9 +438,61 @@ function testAssertObjectEqualsMap() {
...
+  for (var i = 0; i < typedArrays.length; ++i) {
+    var name = typedArrays[i];
+    var arrType = window[name];
+    if (arrType) {
+      function construct(num) { return num; }
+      function mutate(array, offset) { array[0] *= -1; }
+      if (name.startsWith('Big')) {
```
shicks: I'm not in love with all this logic in a test case.  I wonder if simply breaking out a few relevant cases wouldn't be cleaner?

1. same elements in same type of array with different size backing array is equal
2. do a mutation to one of the normal TypedArrays and to one of the BigInt ones and make sure it makes them unequal
3. a simple pair of each are equal (this one should be exhaustive but shouldn't need a for-loop - e.g. "assertObjectEquals(Uint8Array.of(1, 2, 3), Uint8Array.of(1, 2, 3)); assertObjectEquals(Uint16Array.of(1, 2, 3), Uint16Array.of(1, 2, 3));"
4. same elements in different type are unequal (again, just pick a couple pairs)
5. different elements same length => unequal
6. same elements but one extra => unequal

For the bigint ones, we can jam them into a single separate test case where the whole thing is gated on an existence check.

shicks:
Sorry for the slow response - I've been off-rotation the last couple weeks and was assuming someone else would pick things up.  Please take a look at the comments - overall the actual changes look good, it's mainly just some nits on the tests.

shicks:
Internal issue b/131718326 created.

shicks:
Also, please reword the commit message to describe a bit more details of what it's doing, and include a "RELNOTES[new]: assertObjectEquals now checks typed arrays deeply".  Actually, that alone is probably enough.

closure/goog/testing/asserts_test.js
michaelfig: Okay, will do.  (It's my preference... I just got confused by the actual code which seemed to require ES5.)

closure/goog/testing/asserts_test.js
michaelfig: I've done this, which makes all the above review changes irrelevant.  Thanks for helping my write better unit tests!

closure/goog/testing/asserts.js
michaelfig: Done.

michaelfig:
@shicks, squashed and finally passed the checks.

Next time I'll follow the formatting instructions. ;)

shicks:
Look good, thanks - reviewing this internally now.

PR #974, opened 2019-05-20
Fix return type for goog.structs.Trie.getKeyAndPrefixes.

ribrdb:


shicks:
The return type was correct before - Object keys are always strings, and you run into problems if you treat them otherwise.

If you want to pursue the test change we can do that separately.

closure/goog/structs/trie_test.js
```diff
@@ -430,9 +430,12 @@ function testPrefixOptionOnGetKeys() {
 function testGetKeyAndPrefixes() {
   var trie = makeTrie();
   // Note: trie has one of its keys as ''
+  const results = trie.getKeyAndPrefixes('world');
   assertEquals(
       'getKeyAndPrefixes, should be 2', 2,
-      goog.object.getCount(trie.getKeyAndPrefixes('world')));
+      goog.object.getCount(results));
+  assertEquals('an empty string key', results[0]);
+  assertEquals({}, results['world'.length-1]);
```
shicks: What about `assertObjectEquals({0: 'an empty string key', 4: {}}, trie.getKeyAndPrefixes('world'))` ?

shicks:
Created internal issue b/133367790 to track this

shicks:
Tracked by internal b/133368820

ribrdb:
Hmm. So I guess I'm confused because the code in this method is treating the keys as numbers. Are those implicitly converted to strings or something?
What I really want is to find the longest prefix of a string that is contained in the trie.
I added the test because there's no example of how to use the return value of getKeyAndPrefixes.
To find the maximal length prefix I thought I would do something like `Math.max(...goog.object.keys(trie.getKeyAndPrefixes(str))`
Are you saying there's going to be problems with that?

shicks:
That should work, but only because of how JavaScript does implicit coercion between numbers and strings.  All JavaScript Object and Array keys are technically strings (they may be stored as numbers internally to the VM, but this is unobservable) - you can see this by doing a for-in loop over an array (note: this is generally considered a bad idea) and you'll see that the type of the "numbers" you get out is actually string:

```js
const arr = ['a', 'b', 'c'];
for (const i in arr) {
  console.log(`${i} [${typeof i}]`);
}
```

This logs

```
0 [string]
1 [string]
2 [string]
```

But you generally don't run into problems because there's just so much coercion going on.  If you use the index as a number (e.g. by passing it to `Math.max`), it will just parse it and turn it into a number right there.

This is why generally it's better to use the newer ES6 data structures, like `Map`, which actually preserves its keys correctly, without stringifying them all.

shicks:
I'm working on merging a commit to close this internally.

Issue #950, opened 2019-01-09
documentation for goog.dom.createDom is confusing about setting properties vs attributes


ribrdb:
The documentation for `goog.dom.createDom` says it takes "a map of name-value pairs for attributes", but it doesn't actually set attributes. It calls `goog.dom.setProperties()`, which doesn't support custom attributes. The documentation for `createDom` does say "For passing properties, please see `goog.dom.setProperties` for more information."
I understood that to mean "createDom sets attributes, not properties. If you want to set properties you should use `goog.dom.setProperties`," which is pretty much the exact opposite of what's going on.

shicks:
This issue is being closed as a preparation step before archiving the repository.  See the [README](https://github.com/google/closure-library) for more detail.

PR #975, opened 2019-05-20
Clarify documentation for how goog.dom.createDom sets attributes.

ribrdb:
See #950


shicks:
tracked by internal issue b/133368820

PR #982, opened 2019-07-17
Warn for an ignored setter for ParsedDependency#path

koba04:
`ParsedDependency#path` is ignored because the setter is no-op, which makes debugging harder.
So I think it would be nice if there is a warning for the setter.
But warning for every accessing is very annoying, so I limit the warning at once.


nreid260:
Internal Google issue http://b/137758853 created.

koba04:
Thanks!

PR #992, opened 2019-10-24
Check for null when checking typeof websocket opt_params != object

MazinZ:
I'm running into issues with a ClojureScript project that is passing null for opt_params. Since `typeof null` is "object," opt_params doesn't get initialized to the default value. I think the argument can be made that the change should be not to pass null to goog.net.WebSocket, but I think null should explicitly be checked for when testing using typeof and object in a conditional.

### DIFF
--- a/closure/goog/net/websocket.js
+++ b/closure/goog/net/websocket.js
@@ -62,7 +62,7 @@ goog.require('goog.log');
  */
 goog.net.WebSocket = function(opt_params, opt_getNextReconnect) {
   goog.net.WebSocket.base(this, 'constructor');
-  if (typeof opt_params != 'object') {
+  if (typeof opt_params != 'object' || opt_params === null) {
     opt_params = /**@type {!goog.net.WebSocket.Options} */ ({
       autoReconnect: opt_params,
       getNextReconnect: opt_getNextReconnect,


kjin:
Internal Google issue http://b/143291963 created.

closure/goog/net/websocket.js
```diff
@@ -62,7 +62,7 @@ goog.require('goog.log');
  */
 goog.net.WebSocket = function(opt_params, opt_getNextReconnect) {
   goog.net.WebSocket.base(this, 'constructor');
-  if (typeof opt_params != 'object') {
+  if (typeof opt_params != 'object' || opt_params === null) {
     opt_params = /**@type {!goog.net.WebSocket.Options} */ ({
       autoReconnect: opt_params,
```
kjin: This assignment isn't safe when `opt_params === null`.

kjin: Actually I realized that this is OK (but doesn't look great) because the next line forces `this.autoReconnect_` to be boolean. That being said, is it still possible to do what I said in the top-level comment, because technically the type annotations should guard against this already?

MazinZ: Yes, it does slightly differ from convention to do so but I can make a change to avoid passing null here. However, I do think adding a null check here (and really anywhere where typeof "object" comparison is done) makes the intent clearer for additions of future logic.

Either way, I don't feel very strongly about this so feel free to close the issue if you disagree.

kjin: Actually, we are thinking about changing this to only allowing the object parameter, which should simplify this a bit. In this case we just check if the input parameter is falsey or not.

MazinZ: Makes sense!

kjin:
@MazinZ we updated the API to only allow `opt_params`, so I think the root issue that this PR was opened to fix has been addressed. I'll close this PR. Thanks for the contribution and heads up!

PR #993, opened 2019-10-29
Fix disposable leak in field.js

pirosikick:


kjin:
Internal Google issue http://b/143538766 created.

PR #999, opened 2019-11-25
Use copybara to export google3

alexeagle:
This is pushed from the same Base commit as the current closure-library HEAD, so these changes are purely the result of switching from MOE to copybara

PiperOrigin-RevId: 281861339


closure/goog/dom/savedcaretrange_test.js
```diff
@@ -23,7 +23,7 @@ const testingDom = goog.require('goog.testing.dom');
 const userAgent = goog.require('goog.userAgent');
 
 /*
-   TODO(user): Look into why removeCarets test doesn't pass.
+   TODO(ajp): Look into why removeCarets test doesn't pass.
```
shicks: This doesn't look right.

closure/goog/testing/testsuite.js
```diff
@@ -27,13 +27,13 @@ var TestSuiteOptions;
...
  * or goog.modules.
  *
- * @param {!Object<string, function()|!Object>} obj An object with one or more
- *     test methods, and optional setUp, tearDown and getTestName methods. The
- *     object may also have nested Objects (named like tests, i.e.
- *     `testNestedSuite: {}`) that will be treated as nested testSuites. Any
- *     additional setUp will run after parent setUps, any additional tearDown
- *     will run before parent tearDowns. The this object refers to the object
- *     that the functions were defined on, not the full testSuite object.
+ * @param {!Object} obj An object with one or more test methods, and optional
```
shicks: This is its own separate change, caused by the two repos being out of sync

alexeagle: fixed by passing base CL as a bare arg to copybara

doc/develop/conformance_rules.md
```diff
@@ -115,14 +114,12 @@ MyClass.prototype.method = function(arg) {
 }
 ```
 
-
-{: #logger}
-### goog.debug.Logger 
+### goog.debug.Logger {#logger}
```
shicks: This will break our documentation.  The problem is that internal markdown and github-compatible markdown have mutually incompatible syntax for this, so we were using a MOE transformer to fix this.

It's probably fine to take this for now and we can look into fixing it up with some travis post-processing ASAP.

rishipal:
Internal issue created - http://b/145236430

PR #1000, opened 2019-11-29
Fix creating emoji url using id when prefixUrl exists

toshi-toma:
`emojipicker.js#getSelectedEmoji` return `goog.ui.emoji.Emoji` using id when prefixUrl exists.

I checked the [demo page](https://google.github.io/closure-library/source/closure/goog/demos/) of Emoji Picker. This demo doesn't use  urlPrefix_, so working properly. If use urlPrefix_, don't work properly.
https://github.com/google/closure-library/blob/master/closure/goog/demos/popupemojipicker.html#L303

if use `urlPrefix_`, `getSelectedEmoji` returns the `goog.ui.emoji.Emoji` with the ID added to the `urlPreix`. (ex: https://foo/static/std.200)

We expect the `goog.ui.emoji.Emoji` with the URL added to the `urlPreix`(ex: https://foo/static/emoji/200.gif)

### DIFF
--- a/closure/goog/ui/emoji/emojipicker_test.js
+++ b/closure/goog/ui/emoji/emojipicker_test.js
@@ -810,6 +810,30 @@ testSuite({
     picker.dispose();
   },
 
+  testGetSelectedEmoji_urlPrefix() {
+    const defaultImg = 'http://example.com/../../demos/emoji/none.gif';
+    const picker = new EmojiPicker(defaultImg);
+    picker.setDelayedLoad(false);
+    picker.addEmojiGroup(emojiGroup1[0], emojiGroup1[1]);
+    picker.setUrlPrefix('http://example.com/')
+    picker.render();
+
+    const palette = picker.getPage(0);
+
+    // No emoji should be selected yet
+    assertUndefined(palette.getSelectedEmoji());
+
+    // Artificially select the first emoji
+    palette.setSelectedIndex(0);
+
+    // Now we should get the first emoji back. See emojiGroup1 above.
+    const emoji = palette.getSelectedEmoji();
+    assertEquals(emoji.getId(), 'std.200');
+    assertEquals(emoji.getUrl(), 'http://example.com/../../demos/emoji/200.gif');
+
+    picker.dispose();
+  },
+
   testPreLoadCellConstructionForFastLoadingNonProgressive() {
     const defaultImg = '../../demos/emoji/none.gif';
     const picker = new EmojiPicker(defaultImg);
--- a/closure/goog/ui/emoji/emojipicker_test.js
+++ b/closure/goog/ui/emoji/emojipicker_test.js
@@ -811,26 +811,22 @@ testSuite({
   },
 
   testGetSelectedEmoji_urlPrefix() {
-    const defaultImg = 'http://example.com/../../demos/emoji/none.gif';
+    const defaultImg = 'a/b/../../../../demos/emoji/none.gif';
     const picker = new EmojiPicker(defaultImg);
     picker.setDelayedLoad(false);
     picker.addEmojiGroup(emojiGroup1[0], emojiGroup1[1]);
-    picker.setUrlPrefix('http://example.com/')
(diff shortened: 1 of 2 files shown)


blickly:
Google Internal issue http://b/145526555 created 

(Also, I think that the bot is asking you to sign a CLA by going to https://cla.developers.google.com/)

toshi-toma:
YES!! I'll check CLA.

blickly:
Still waiting on CLA, AFAICT

toshi-toma:
@googlebot I signed it!

toshi-toma:
@googlebot I signed it!

12wrigja:
Thanks for contributing! 

This fix looks great, but we'd like to see a test that utilizes the URL prefix so this doesn't regress. Something similar to the existing test for getSelectedEmoji should work: https://github.com/google/closure-library/blob/master/closure/goog/ui/emoji/emojipicker_test.js#L734

toshi-toma:
@12wrigja 
I added a test and apply clang-format.
but CI is failed...

closure/goog/ui/emoji/emojipicker_test.js
```diff
@@ -810,6 +810,31 @@ testSuite({
...
+    const palette = picker.getPage(0);
+
+    // No emoji should be selected yet
+    assertUndefined(palette.getSelectedEmoji());
+
+    // Artificially select the first emoji
+    palette.setSelectedIndex(0);
+
+    // Now we should get the first emoji back. See emojiGroup1 above.
+    const emoji = palette.getSelectedEmoji();
```
12wrigja: This should be the picker object, not the underlying palette, or your changes aren't called.

As the picker's selected emoji is updated via action events (from the palettes it contains), you'll also need to dispatch an ACTION event from the palette to get the picker's getSelectedEmoji to work:

```
// Artificially select the first emoji
palette.setSelectedIndex(0);
palette.dispatchEvent(goog.ui.Component.EventType.ACTION);

// Now we should get the first emoji back. See emojiGroup1 above.
const emoji = picker.getSelectedEmoji();
```

closure/goog/ui/emoji/emojipicker_test.js
```diff
@@ -810,6 +810,31 @@ testSuite({
...
+    assertEquals(emoji.getId(), 'std.200');
+    assertEquals(
+        emoji.getUrl(), 'http://example.com/../../demos/emoji/200.gif');
```
12wrigja: For both of these assertEquals calls, the first value is the expected value, and the second the actual. Please swap. The other test in this file is wrong too, but I'll fix that internally.

closure/goog/ui/emoji/emojipicker_test.js
```diff
@@ -810,6 +810,31 @@ testSuite({
     picker.dispose();
   },
 
+  testGetSelectedEmoji_urlPrefix() {
+    const defaultImg = 'http://example.com/../../demos/emoji/none.gif';
+    const picker = new EmojiPicker(defaultImg);
+    picker.setDelayedLoad(false);
+    picker.addEmojiGroup(emojiGroup1[0], emojiGroup1[1]);
+    picker.setUrlPrefix('http://example.com/')
```
12wrigja: This URL prefix may not work, as that file doesn't exist. I'd recommend instead using a prefix that is really just adding useless pieces to the path, e.g. 'a/b/../../' that is in effect a no-op when loaded by the browser but will cause the underlying URL to look different.

12wrigja: The CI seems to fail because you are missing a ; here, and the CI formatter wants to move the previous line onto this line.

shicks: We'll fix this internally.

toshi-toma: OK!!

toshi-toma:
@12wrigja 
Thank you for checking and review.
I fixed those.

PR #1007, opened 2019-12-18
fix ui/popupbase.js

sakito21:
Ignore deactivate events from partner elements in IE.


sakito21:
https://bozuman.cybozu.com/k/#/space/3709/thread/32176/2190298

kjin:
Internal issue http://b/146525538 created

PR #1008, opened 2019-12-26
Fix deleteW3cRange_ in enterhanderjs

toshi-toma:
Some behavior of deleteW3cRange_ is inadequate.
This behavior only happens using `Chrome`  or `Safari`.

<details><summary>concrete example</summary>

In the case of Editor Demo. `|` is the cursor.
https://google.github.io/closure-library/source/closure/goog/demos/editor/editor.html

```
a
|b|
c
d
```

type Enter...

```
a

|cd
```

Other

```
a
|b
c|c
d
```

type Enter...

```
a

|cd
```

Other

```
|a|
b
c
d
```

type Enter...

```
b
c
d
|
```
</details>

Fixes
- Normalize range in `isPartialEndW3c_ `, Since the range object is different for each browser
- Concatenate characters only when `isPartialStart` and `isPartialEnd`
- When the selection starts from the beginning, the Caret position is at the beginning

### DIFF
--- a/closure/goog/editor/plugins/enterhandler.js
+++ b/closure/goog/editor/plugins/enterhandler.js
@@ -631,6 +631,22 @@ goog.editor.plugins.EnterHandler.prototype.deleteCursorSelectionW3C_ =
   return goog.editor.range.getDeepEndPoint(range, true);
 };
 
+/**
+ * Checks Whether the selection range start from beginning.
+ * @param {Node} node the element or text node the range starts in.
+ * @param {Node} baseNode the container element.
+ * @return {boolean} Whether the selection range start from beginning.
+ * @private
+ */
+goog.editor.plugins.EnterHandler.isStartFromBeginning_ = function(node, baseNode) {
+  while (node && node.nodeName != goog.dom.TagName.BODY && node != baseNode) {
+    if (node.previousSibling) {
+      return false;
+    }
+    node = node.parentNode;
+  }
+  return true;
+};
 
 /**
  * Deletes the contents of the selection from the DOM.
@@ -655,16 +671,22 @@ goog.editor.plugins.EnterHandler.deleteW3cRange_ = function(range) {
     var isPartialEnd = !isInOneContainer &&
         goog.editor.plugins.EnterHandler.isPartialEndW3c_(range);
 
+    var isStartFromBeginning =
+        goog.editor.plugins.EnterHandler.isStartFromBeginning_(
+            range.getStartNode(), baseNode);
+
     // Remove The range contents, and ensure the correct content stays selected.
     range.removeContents();
     var node = nodeOffset.findTargetNode(baseNode);
     if (node) {
       range = goog.dom.Range.createCaret(node, rangeOffset);
     } else {
-      // This occurs when the node that would have been referenced has now been
-      // deleted and there are no other nodes in the baseNode. Thus need to
-      // set the caret to the end of the base node.
-      range = goog.dom.Range.createCaret(baseNode, baseNode.childNodes.length);
+      // when the node that would have been referenced has now been deleted and
+      // there are no other nodes in the baseNode,  Thus need to set the caret
+      // to the end of the base node. If selection range start from beginning,
+      // set the caret to the start of the base node.
+      var pos = isStartFromBeginning ? 0 : baseNode.childNodes.length;
+      range = goog.dom.Range.createCaret(baseNode, pos);
       reselect = false;
     }
     range.select();
--- a/closure/goog/editor/plugins/enterhandler.js
+++ b/closure/goog/editor/plugins/enterhandler.js
@@ -649,6 +649,8 @@ goog.editor.plugins.EnterHandler.deleteW3cRange_ = function(range) {
     var isInOneContainer =
         goog.editor.plugins.EnterHandler.isInOneContainerW3c_(range);
 
+    // Whether the selection starts in a container.
+    var isPartialStart = !isInOneContainer && range.getStartOffset() != 0;
     // Whether the selection ends in a container it doesn't fully select.
     var isPartialEnd = !isInOneContainer &&
         goog.editor.plugins.EnterHandler.isPartialEndW3c_(range);
@@ -683,7 +685,7 @@ goog.editor.plugins.EnterHandler.deleteW3cRange_ = function(range) {
       }
     }
 
-    if (isPartialEnd) {
+    if (isPartialStart && isPartialEnd) {
       /*
        This code handles the following, where | is the cursor:
          <div>a|b</div><div>c|d</div>
--- a/closure/goog/editor/plugins/enterhandler.js
+++ b/closure/goog/editor/plugins/enterhandler.js
@@ -757,6 +757,21 @@ goog.editor.plugins.EnterHandler.isInOneContainerW3c_ = function(range) {
 goog.editor.plugins.EnterHandler.isPartialEndW3c_ = function(range) {
   var endContainer = range.getEndNode();
   var endOffset = range.getEndOffset();
+  // Since the range object is different for each browser,
+  // Normalize range using previousSibling or parentNode of
+  // endContainer, when endOffset is 0.
+  while (endOffset == 0 && endContainer) {
+    if (endContainer.previousSibling) {
+      endContainer = endContainer.previousSibling;
+      endOffset = goog.editor.node.getLength(endContainer);
+    } else if (endContainer.parentNode) {
+      endContainer = endContainer.parentNode;
+      endOffset = 0;
+    } else {
+      break;
+    }
+  }
+
   var node = endContainer;
   if (goog.editor.style.isContainer(node)) {
     var child = node.childNodes[endOffset];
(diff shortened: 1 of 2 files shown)


shicks:
Wow, that's really something!  Thanks so much for contributing this fix!

shicks:
Oops - that was the wrong label to have our bot pick it up.  Fixed the label and deleted the broken one so we don't make that mistake again.

shicks:
This is pulled in internally as cl/299996311

closure/goog/editor/plugins/enterhandler.js
```diff
@@ -757,6 +781,21 @@ goog.editor.plugins.EnterHandler.isInOneContainerW3c_ = function(range) {
...
+    if (endContainer.previousSibling) {
+      endContainer = endContainer.previousSibling;
+      endOffset = goog.editor.node.getLength(endContainer);
+    } else if (endContainer.parentNode) {
+      endContainer = endContainer.parentNode;
+      endOffset = 0;
+    } else {
+      break;
+    }
+  }
```
12wrigja: Could this be replaced with a call to "goog.editor.range.normalize"?

closure/goog/editor/plugins/enterhandler.js
```diff
@@ -631,6 +631,22 @@ goog.editor.plugins.EnterHandler.prototype.deleteCursorSelectionW3C_ =
...
 };
 
+/**
+ * Checks Whether the selection range start from beginning.
+ * @param {Node} node the element or text node the range starts in.
+ * @param {Node} baseNode the container element.
+ * @return {boolean} Whether the selection range start from beginning.
+ * @private
+ */
+goog.editor.plugins.EnterHandler.isStartFromBeginning_ = function(node, baseNode) {
```
12wrigja: This method took me a bit to understand - can we think of a better name here that is more descriptive? Maybe isNodeLeftMostChild ?

12wrigja:
@toshi-toma Any updates here?

toshi-toma:
@12wrigja Sorry for late. I'll check the comments and fix them.

toshi-toma:
> does it matter which order you make the selections in? I'm sometimes unable to reproduce your examples, and I'm wondering if part of that is because of the order I'm making the selection.

Probably not...
it happens every time in my environment.

12wrigja:
Ah - thanks for clarifying :) It does indeed reproduce now. Thanks. I'll take another look over your changes now that I can repro this and see if there's anything else I have to comment on.

closure/goog/editor/plugins/enterhandler.js
toshi-toma: OK! I'll check it.

12wrigja:
Hi @toshi-toma,

We'd really like to see this get merged in before next week. Is there any help I can give you to move you forward? I don't have any further comments for your code that is currently on Github.

toshi-toma:
@12wrigja 

Thanks for the review. and Sorry for couldn't fix your comments.
Now, I'll fix
- add test
- replaced with a call to "goog.editor.range.normalize"? (might not be)
- rename `isStartFromBeginning_ `

Is this understanding of mine correct?

12wrigja:
Yep! If you only have a limited amount of time to work on this, the thing we need is tests before we can try and merge this :)

closure/goog/editor/plugins/enterhandler.js
toshi-toma: It seems can replace with a call to "goog.editor.range.normalize".
but, The result wasn't to be the expected return value...

closure/goog/editor/plugins/enterhandler.js
toshi-toma: OK!! I'll rename it to `isNodeLeftMostChild `.

toshi-toma: 094d698

toshi-toma:
I don't know how to run unit-tests in local.
Where should I look?

12wrigja:
Ah sorry - I was mislead thinking that there was a workflow for this externally. Seems there isn't really one (you have to generate various files by hand and then manually execute the HTML in a browser, nothing nice like `npm test` :( ).

What we can do is once you've finished with your commits I can import them internally, write the unit tests for this code, submit it, and sync this change back out. I've been told that you will correctly get attribution for all the commits you've done, and all the test code will be attributed to me. Sound reasonable?

toshi-toma:
Yep! It sounds like a good idea :)
So, Is there something missing for me?

closure/goog/editor/plugins/enterhandler.js
12wrigja: That's fine. I'll take a closer look at this internally and change it if there's something similar we can use instead.

12wrigja:
I don't think so! Thanks. I'll try and merge this in by the end of this week.

closure/goog/editor/plugins/enterhandler.js
toshi-toma: Thanks！！

toshi-toma:
Thanks!!

12wrigja:
Hi @toshi-toma ,

I figured out how to generate the test pages (HTML pages) and added in an npm script to both generate the html and another to serve it from the right directory on a personal fork: https://github.com/12wrigja/closure-library/tree/enterhandler_with_tests (see package.json and the commit description). I'll try and get these upstreamed into the library proper, but that commit can serve  as a stopgap. I also spent some time creating some unit tests based on your examples and put them in the right file.

Unfortunately, based on the examples and the test I wrote, I can't seem to get the tests to pass without changing your code to account for more complex cases (e.g. does the selected range being replaced with a newline encompass multiple "lines" in the editor?). I don't have time to work on solving these cases myself (I'm actually quite unfamiliar with the Editor package) - are you interested in taking some time to fix up your code?

How did you test your changes in the first place? By running the demos somehow?

toshi-toma:
Thanks!!
I'm trying to test with your repo.

toshi-toma:
> How did you test your changes in the first place? By running the demos somehow?

I checked by running the demos!

> Unfortunately, based on the examples and the test I wrote, I can't seem to get the tests to pass without changing your code to account for more complex cases

`deleteW3cRange_` doesn't perform to insert a new line.

- `testDeleteW3CRemoveEntireFirstLineWithShiftDown`
  -  I think it expect `<div>b</div><div>c</div><div>d</div>` at `deleteW3cRange_`
  -  It seems insert <br> at later.
- `testDeleteW3CRemoveEntireLineWithPartialSecondLine`
  - I think It expect `<div>a</div><div>c</div><div>d</div>`
- `testDeleteW3CRemoveEntireLineWithShiftDown`
  - I think it expect `<div>a</div><div>c</div><div>d</div>`
  -  It seems insert <br> at later.

closure/goog/editor/plugins/enterhandler.js
```diff
@@ -245,7 +245,7 @@ goog.editor.plugins.EnterHandler.prototype.deleteBrGecko = function(e) {
         //    end of that.
         // 2. If the BR doesn't have a previous sibling or the previous sibling
         //    is a block level element or a BR, we place the cursor at the
-        //    beginning of the leftmost leaf of its next sibling.
+        //    leftmost of the leftmost leaf of its next sibling.
```
12wrigja, line 248: I don't think this should have changed - this comment seems accurate as it was.

12wrigja:
[It does sometimes](https://github.com/google/closure-library/blob/755f721238c8349e8c18fd23d725cb65075fca64/closure/goog/editor/plugins/enterhandler.js#L664-L676) which is what confused me. Turns out the missing newline is probably from the browser itself (Chrome will add <div><br></div> when enter is pressed on a contenteditable div by default). I don't see a great way to trigger this behavior from the unit tests, so for the moment it makes sense to me to:
 - add some comments to each test indicating what is seen in an actual editor in Chrome (with the extra br's)
 - adjust the unit tests to match what the test is expecting.

If you make those small changes, then I think this is good to be merged. Sorry this has been so hard, and thank you for your patience.

closure/goog/editor/plugins/enterhandler.js
toshi-toma, line 248:
```suggestion
        //    beginning of the leftmost leaf of its next sibling.
```

toshi-toma: Thanks!!

d0cec01

toshi-toma:
> so for the moment it makes sense to me to:

I'm with you. Thanks!!
Is there something missing for me?

12wrigja:
If the tests pass for you, then you should add them into your PR and add the relevant comments (as discussed above). I think that's it - thanks for resolving the other comment.

12wrigja:
@toshi-toma Any updates here? Also, I thought you had signed the CLA?

toshi-toma:
Could I add the tests from your [branch](https://github.com/12wrigja/closure-library/tree/enterhandler_with_tests)??

12wrigja:
Yes you can add those, although you'll need to change them to pass (as I haven't changed them since I first wrote them and they were failing then).

toshi-toma:
@googlebot I fixed it.

toshi-toma:
@12wrigja 
I added some tests and update CLA!!
But Now, CI failed by format...

I install clang and ran format scripts. but It doesn't work properly...

12wrigja:
Don't worry about the formatting - I'll fix it internally. I think this looks good otherwise - I've sent it for review internally.

12wrigja:
Thanks for the patience and hard work, @toshi-toma! This is now merged.

toshi-toma:
@12wrigja 
Thank you so much!!

PR #1010, opened 2020-01-04
Move promises-aplus-tests from deps to devDeps

teppeis:
`promises-aplus-tests` is used only in `goog/promise/testsuiteadapter.js` that is not used in production.
It should be moved to devDependencies.

https://github.com/google/closure-library/blob/36d2c6df96b21043361b3bca4b7b5e981cb91471/closure/goog/promise/testsuiteadapter.js#L33

https://www.npmjs.com/package/google-closure-library?activeTab=dependencies

Also this PR removes `goog/promise/testsuiteadapter.js` from published npm package.
It is obviously a development-only file, as it is excluded from type checking.

https://github.com/google/closure-library/blob/36d2c6df96b21043361b3bca4b7b5e981cb91471/scripts/ci/compile_closure.sh#L41

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,6 +6,7 @@
 /*.json
 /closure/bin/generate_closure_unit_tests
 /closure/goog/demos/
+/closure/goog/promise/testsuiteadapter.js
 /closure-deps/
 /doc
 /scripts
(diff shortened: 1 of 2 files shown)


kjin:
Internal issue created: http://b/147245657

kjin:
This is landed in 196ccdb950f8894a0a50dae7c20bb587bcdbb2f5

PR #1039, opened 2020-02-06
Allow passing Uint8Array to StreamParser.parse

travikk:
This will help address issue with grpc-web described here: https://github.com/grpc/grpc-web/pull/683#issuecomment-561999180

In a nutshell, grpc-web implements StreamParser which accepts only Array or ArrayBuffer (as per the interface), but the upstream to that parser operates on UInt8Array. For various reasons (described in the link above), the upstream creates a copy of all the bytes first before passing to the parser.
this obviously is CPU and memory ineffective and for larger payloads (50+ mb) creates a browser freeze.

We can solve those problems by passing UInt8Array to the parser, but first the interface signature has to be addressed, hence this PR.

### DIFF
--- a/closure/goog/net/streams/jsonstreamparser.js
+++ b/closure/goog/net/streams/jsonstreamparser.js
@@ -239,7 +239,7 @@ Parser.prototype.getExtraInput = function() {
 
 
 /**
- * @param {string|!ArrayBuffer|!Array<number>} input
+ * @param {string|!ArrayBuffer|!Uint8Array|!Array<number>} input
  *     The current input string (always)
  * @param {number} pos The position in the current input that triggers the error
  * @throws {!Error} Throws an error indicating where the stream is broken
--- a/closure/goog/net/streams/streamparser.js
+++ b/closure/goog/net/streams/streamparser.js
@@ -57,7 +57,7 @@ goog.net.streams.StreamParser.prototype.getErrorMessage = goog.abstractMethod;
  *
  * Note that there is no Parser state to indicate the end of a stream.
  *
- * @param {string|!ArrayBuffer|!Array<number>} input The input data
+ * @param {string|!ArrayBuffer|!Uint8Array|!Array<number>} input The input data
  * @throws {!Error} if the input is invalid, and the parser will remain invalid
  *    once an error has been thrown.
  * @return {?Array<string|!Object>} any parsed objects (atomic messages)


travikk:
@googlebot I signed it!

shicks:
Thanks for the PR.

There's also (potentially more relevant) assertions in [PbStreamParser](https://github.com/google/closure-library/blob/20989980d5922c143c019a84f2610e8862826718/closure/goog/net/streams/pbstreamparser.js#L186), as well as one we'd need to deal with in some internal gRPC code (no need to worry about that one).

But more importantly, what is your purpose for doing this?  It does already "accept" an `ArrayBuffer`, notwithstanding the fact that all the implementations seem to just immediately assert on a subtype of the parameter type, basically invalidating the whole interface contract in the first place.  But that means that any existing implementations aren't going to handle Uint8Array anyway (even if they should be able to).  I'm open to accepting this, but want to get a better understanding of the motivation.  Thanks!

closure/goog/net/streams/jsonstreamparser.js
```diff
@@ -239,7 +239,7 @@ Parser.prototype.getExtraInput = function() {
 
 
 /**
- * @param {string|!ArrayBuffer|!Array<number>} input
+ * @param {string|!ArrayBuffer|!Uint8Array|!Array<number>} input
```
shicks: This seems unnecessary - it should be fine as simply {string}, since parse() immediately asserts that the input is a string - what directed you to make this change?  If you're seeing a type checking error then it sounds like there's a bug somewhere else.

travikk:
@shicks excellent question. The reason why I'm adding Uint8Array as allowed parameter is to allow grpc-web parser to parse it. 

If you take a look at this guy here: https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpcwebclientreadablestream.js#L140 it passes a copy of `byteSource` which is the underlaying issue I'm trying to address.

`[].slice.call(byteSource)` will result in memory copy of all the bytes inside the underlaying ArrayBuffer and return a signature of Array, which is why the type check now works. My solution is to change  that line to the following:

```
self.parser_.parse([].slice.call(byteSource))
```

but this isn't going to work, as byteSource is Uint8Array. Type check will fail, as the interface now doesn't allow that.

Now you could say to pass the underlaying ArrayBuffer directly into the parser and you'd be right - this was my first proposed solution. There is an issue however with that, as `googCrypt.decodeStringToUint8Array` (line 133 in referenced file) initially allocates more bytes that the actual data has. That results in some internal-to-google tests failing.

this brings me to my conclusion - pass the view of the ArrayBuffer which knows what is the correct size of the buffer.

More details here: https://github.com/grpc/grpc-web/pull/683

closure/goog/net/streams/jsonstreamparser.js
travikk: Agreed. Reason is, `error_` is invoked from the `parse` method, which now obviously accepts UInt8Array.
You're right in saying that the first things `parse` does is it asserts the type to a string, but that doesn't change it's type signature, I think a cast would be required and then the signature of `error_` could be stripped down to just string.

...is this flow you guys are using here?

shicks: No, we use [Closure Compiler](https://github.com/google/closure-compiler).  It recognizes goog.asserts.assertString as narrowing the type of its argument to be only string, which is why I was confused about where you were actually seeing the error.  What type checker are you using?

travikk: @shicks it was the Travis build that failed. Can't find the failed build in the history though.

shicks:
I suspected as much... I commented on the grpc bug since I believe the proposal to simply change the accepted type in the subclass was rejected incorrectly.

shicks:
So should I close this PR, then?  If you still think it makes sense in light of the grpc changes, please reopen and we'll take another look.

PR #1063, opened 2020-03-06
Fix grammatical mistake in JSDocs for `goog.html.SafeUrl.getTypedStringValue` and `goog.string.Const.getTypedStringValue`

DCtheTall:
Both JSDoc comments said:

"Returns this ... value a string"

when it should be

"Returns this ... value as a string"


12wrigja:
This was merged to master with a36bf4d9760185b44b6bcf8c1f745cd0b7ed6977, unsure why this hasn't been closed. Probably something wrong in the externalization workflow.

PR #1080, opened 2020-07-07
Convert goog.events.BrowserFeature to goog.module

CarlosM10011:
When compiling something that depends on this file, the closure compiler will give me an internal compiler error. I figured that since lines from this file were present in the  stack trace, I could try to get rid of the `goog.scope` and make this file use `goog.module`. The error appears to be gone now.

I am not sure if different builds of the compiler fix this error, but if it helps, I am using a maven plugin to run it located at (https://github.com/blutorange/closure-compiler-maven-plugin).


12wrigja:
Would you mind posting the full error you get from the compiler?

We can try and land this change, but it might be hard to land this internally. I'll pull it in and see.

12wrigja:
There are some internal build failures that need to be fixed before this can land - once they have done so I'll attempt to get this merged. Nothing for you to do now but wait.

CarlosM10011:
Sounds good.

CarlosM10011:
H

CarlosM10011:
I just fixed the clang-format errors; I was using a different one in my path.

12wrigja:
This has been merged. Thanks for contributing!

PR #1088, opened 2020-09-29
Experiment with what it might look like to check Clutz/Tsickle output into repo

shicks:


shicks:
We've got a better approach with only a single file.

PR #1103, opened 2021-01-07
[closure-deps] Allow globs in --exclude flag

kjin:
This change adds the ability to specify globs in closuremakedeps's --exclude flag, instead of just path prefixes.

### DIFF
--- a/closure-deps/bin/closuremakedeps.js
+++ b/closure-deps/bin/closuremakedeps.js
@@ -24,6 +24,7 @@
 const depFile = require('../lib/depfile');
 const depGraph = require('../lib/depgraph');
 const fs = require('fs');
+const minimatch = require('minimatch');
 const parser = require('../lib/parser');
 const path = require('path');
 const process = require('process');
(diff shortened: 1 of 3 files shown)


closure-deps/bin/closuremakedeps.js
```diff
@@ -52,14 +53,15 @@ function parseArgs(args) {
...
-            'Closure Library\'s deps.js file, then there is no need to ' +
-            'include closure/goog/array/array.js as an input file as well. ' +
+        description:
+            'One or more input files to calculate dependencies for. The ' +
+            'namespaces in these files will be combined with those given ' +
+            'with the --root flag to form the set of namespaces to find ' +
+            'dependencies for. If this is a file that contains repeated ' +
+            'calls to goog.addDependency, then the calls will be used as ' +
+            'forward declarations. e.g. if you include Closure Library\'s ' +
+            'deps.js file, then there is no need to include ' +
```
12wrigja, line 63: Seems unrelated - revert?

closure-deps/bin/closuremakedeps.js
```diff
@@ -215,7 +226,7 @@ async function main(opt_args) {
 
   const sources = new Set((args.file || []).map(resolve));
   const roots = (args.root || []).map(resolve);
-  const excluded = new Set((args.exclude || []).map(resolve));
```
12wrigja, line 218: Doesn't this still make sense to be treated as a set? e.g. if someone specifies `--exclude=path/to/*_test.js --exclude=path/to/*_test.js` we should only process that glob once?

closure-deps/bin/closuremakedeps.js
```diff
@@ -215,7 +226,7 @@ async function main(opt_args) {
...
+  const excluded = (args.exclude || []).map(resolve);
```
12wrigja, line 229: This tries to resolve a string as a path relative to the cwd - this doesn't make sense if these are now globs?

closure-deps/bin/closuremakedeps.js
```diff
@@ -129,11 +140,11 @@ function promisify(fn, ...args) {
 
 /**
  * @param {string} pathToScan
- * @param {!Set<string>} excluded
+ * @param {!Array<string>} excluded
```
12wrigja, line 143: Nit: maybe name this excludedGlobs so it's clearer these are not file path references.

12wrigja (changes requested):

closure-deps/bin/closuremakedeps.js
kjin: Yep, trying to figure out why Travis's clang-format is not cooperating with my local one

Issue #1078, opened 2020-06-10
goog.net.WebSocket.EventType.CLOSED does not include any information from the underlying ClosedEvent


drewkutilek:
`goog.net.WebSocket.CLOSED` is triggered as an event without any payload. The underlying `ClosedEvent` has useful information about why the WebSocket was closed, including `code` and `reason`.

https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent

Including a `goog.net.WebSocket.ClosedEvent` to wrap that data would be useful, especially if the server specific implementation relies on the `code` and `reason` to notify the user of a specific error and display a message.

```
/**
 * Object representing a closed event.
 *
 * @param {number} code
 * @param {string} reason
 * @param {boolean} wasClean
 * @extends {goog.events.Event}
 * @constructor
 * @final
 */
goog.net.WebSocket.ClosedEvent = function(code, reason, wasClean) {
  goog.net.WebSocket.ClosedEvent.base(
      this, 'constructor', goog.net.WebSocket.EventType.CLOSED);

  /**
   * @type {number}
   */
  this.code = code;

  /**
   * @type {string}
   */
  this.reason = reason;

  /**
   * @type {boolean}
   */
  this.wasClean = wasClean;
};
goog.inherits(goog.net.WebSocket.ClosedEvent, goog.events.Event);
```

And dispatching the event in `onClose_` like this:

```
/**
 * Called when the web socket has closed.
 *
 * @param {!Event} event The close event.
 * @private
 */
goog.net.WebSocket.prototype.onClose_ = function(event) {
  goog.log.info(this.logger_, 'The WebSocket on ' + this.url_ + ' closed.');

  // Firing this event allows handlers to query the URL.
-  this.dispatchEvent(goog.net.WebSocket.EventType.CLOSED);
+ this.dispatchEvent(new goog.net.WebSocket.ClosedEvent(event.code, event.reason, event.wasClean));

  // Always clear out the web socket on a close event.
  this.webSocket_ = null;

  // See if this is an expected call to onClose_.
  if (this.closeExpected_) {
    goog.log.info(this.logger_, 'The WebSocket closed normally.');
    // Only clear out the URL if this is a normal close.
    this.url_ = null;
    this.protocol_ = undefined;
  } else {
    // Unexpected, so try to reconnect.
    goog.log.error(
        this.logger_, 'The WebSocket disconnected unexpectedly: ' + event.data);

    // Only try to reconnect if it is enabled.
[...]

PR #1119, opened 2021-03-02
modified goog.net.WebSocket.EventType.CLOSED

drewkutilek:
modified goog.net.WebSocket.EventType.CLOSED to include information from the underlying ClosedEvent

https://github.com/google/closure-library/issues/1078

### DIFF
--- a/closure/goog/net/websocket.js
+++ b/closure/goog/net/websocket.js
@@ -27,6 +27,7 @@
  */
 
 goog.provide('goog.net.WebSocket');
+goog.provide('goog.net.WebSocket.ClosedEvent');
 goog.provide('goog.net.WebSocket.ErrorEvent');
 goog.provide('goog.net.WebSocket.EventType');
 goog.provide('goog.net.WebSocket.MessageEvent');
@@ -405,7 +406,7 @@ goog.net.WebSocket.prototype.onClose_ = function(event) {
   goog.log.info(this.logger_, 'The WebSocket on ' + this.url_ + ' closed.');
 
   // Firing this event allows handlers to query the URL.
-  this.dispatchEvent(goog.net.WebSocket.EventType.CLOSED);
+  this.dispatchEvent(new goog.net.WebSocket.ClosedEvent(event.code, event.reason, event.wasClean));
 
   // Always clear out the web socket on a close event.
   this.webSocket_ = null;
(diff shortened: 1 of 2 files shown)


drewkutilek:
@googlebot I signed it!
________________________________

drewkutilek:
@googlebot I signed it!

drewkutilek:
https://cla.developers.google.com/clas does not have the agreement listed for my G Suite account (`kutid002@apps.disney.com`) which should be a member of the google group `ma-google-open-source@apps.disney.com`. Perhaps it will take some time for the signed agreement to be processed. I will check that page periodically and re-run the CLA check once it is there.

kjin:
Hi Drew, how long has it been since the corporate CLA went through? According to the internal guidance regarding CLAs, it might take a few days after the corporate CLA is set up for our database to be updated. So it's possible if it was just set up recently that the database hasn't been updated yet.

I do see an entry for Disney Streaming Services, but I can't tell how long ago the CLA was added -- do you happen to be under this subsidiary?

drewkutilek:
It was just submitted and DocuSign'd this morning. So I will give it a few days and try again.

I do not fall under Disney Streaming Services, I submitted a new one for Movies Anywhere LLC.

kjin:
Ok, I will also check back over the next few days!

drewkutilek:
@googlebot I signed it!

closure/goog/net/websocket.js
```diff
@@ -27,6 +27,7 @@
  */
 
 goog.provide('goog.net.WebSocket');
+goog.provide('goog.net.WebSocket.ClosedEvent');
```
kjin, line 30: Just a heads up -- We will likely delete all of the goog.provides except for `goog.net.WebSocket` at some point in the future (probably within the next year. If you are requiring this in a `goog.provide` file, you should simply require `goog.require('goog.net.WebSocket')` to make it forward-compatible; if you're requiring this in a `goog.module` file, you would need to import it as `const {ClosedEvent} = goog.require('goog.net.WebSocket')`.

kjin (approved):
Thanks! We will merge this internally.

closure/goog/net/websocket.js
kjin: As an additional heads up, we will actually be removing all of the provides (except for the first one) in this change as well.

PR #1135, opened 2021-07-23
Introduce asynchronous MockClock configuration & Async only MockClock

vartan:
Introduces a new configuration object that can be passed to the MockClock constructor. When the MockClock uses the legacy constructor API (either no argument or a boolean argument), it will have the same behavior as before this change (`{synchronous:true, resetScheduler: false, autoInstall}`). 

When MockClock is configured to run asynchronously (`new MockClock({})` or `new MockClock({synchronous: false})`), MockClock will no longer call `goog.async.run.forceNextTick` or replace `setImmediate` when installed. Additionally, the synchronous `tick` and `tickPromise` APIs will no longer be available.  Users should instead call `await tickAsync(millis)` or `await tickAndResolve(millis, promise)` to advance the clock, which allows the browser to return to the event queue and flush any pending work on the event queue.

Additionally, MockClock can be configured with `{synchronous: true, resetScheduler: true}` to use the synchronous behavior but restore `goog.async.run.schedule_` to use the real scheduler when uninstalled instead of forcing next tick and causing side effects for the rest of tests.

### DIFF
--- a/closure/goog/async/run.js
+++ b/closure/goog/async/run.js
@@ -115,6 +115,14 @@ if (goog.DEBUG) {
     goog.async.run.workQueueScheduled_ = false;
     goog.async.run.workQueue_ = new goog.async.WorkQueue();
   };
+
+
+  /**
+   * Resets the scheduler. Only available for tests in debug mode.
+   */
+  goog.async.run.resetScheduler = function() {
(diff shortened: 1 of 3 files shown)


vartan:
going to try to submit this internally instead.

PR #1145, opened 2021-10-08
Harmonize currency format for DACH region

### LINKED ISSUE #1144: Austrian (de-AT) currency format is defaulting to a barely used format
https://github.com/google/closure-library/blob/61eef0f7ac67618a504c976a0bdb4f124f7eaa86/closure/goog/i18n/numberformatsymbols.js#L600

Usually the preferred way to display currency string is that we put the currency code at the end of the text. (Similar as for Germany)
The current pattern makes Angular create really strange currency strings when using [formatCurrency](https://angular.io/api/common/formatCurrency)(-10, 'de-AT', 'EUR', 'EUR'). It will result in "-EUR 10,00", which is really not common here and seen as a bug by the majority of the people.

From my perspective the patterns may be checked by citizens of the according countries. I suggest that the currency pattern is changed to this format for whole DACH region (Germany, Austria, Switzerland).

CURRENCY_PATTERN: '#,##0.00 ¤;-#,##0.00 ¤'

simonbraeuer:
Apply the most common format for currency strings to "de", "de-AT", "de-CH". 
First sign, then value, then currency symbol/code.

Fixes #1144

### DIFF
--- a/closure/goog/i18n/numberformatsymbols.js
+++ b/closure/goog/i18n/numberformatsymbols.js
@@ -576,7 +576,7 @@ goog.i18n.NumberFormatSymbols_de = {
   DECIMAL_PATTERN: '#,##0.###',
   SCIENTIFIC_PATTERN: '#E0',
   PERCENT_PATTERN: '#,##0 %',
-  CURRENCY_PATTERN: '#,##0.00 ¤;-#,##0.00 ¤',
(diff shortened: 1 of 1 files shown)


12wrigja:
These files are internally generated at Google from CLDR data - any changes we make here will be overwritten by the next CLDR data push. This needs to be fixed upstream in order to "stick".

12wrigja:
Closing as we can't accept changes to these files this way. Please follow along on your other issue for resolution.

PR #1146, opened 2021-10-14
Fix a comment in heap.js

sukolsak:
In a min heap, where smaller keys rise to the top, the up-heap operation should move the parent of a node down if the parent is _greater_ than the node.


sukolsak:
@googlebot I signed it!

PR #1158, opened 2022-03-25
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/closure/goog/dom/browserrange/abstractrange.js
+++ b/closure/goog/dom/browserrange/abstractrange.js
@@ -257,7 +257,7 @@ goog.dom.browserrange.AbstractRange.prototype.getHtmlFragment = function() {
           output.append(html);
         } else {
           var index = html.lastIndexOf('<');
-          output.append(index ? html.substr(0, index) : html);
+          output.append(index !== -1 ? html.slice(0, index) : html);
         }
       }
     }
--- a/closure/goog/dom/browserrange/abstractrange.js
+++ b/closure/goog/dom/browserrange/abstractrange.js
@@ -257,7 +257,12 @@ goog.dom.browserrange.AbstractRange.prototype.getHtmlFragment = function() {
           output.append(html);
         } else {
           var index = html.lastIndexOf('<');
-          output.append(index !== -1 ? html.slice(0, index) : html);
+          // if index is -1, then this appends nothing.
+          // if index is 0, then the entire HTML content should be added.
+          // if the index is > 0, then only the portion of the html before the last open tag is appended.
+          if (index !== -1) {
+            output.append(index > 0 ? html.slice(0, index) : html);
+          }
         }
       }
     }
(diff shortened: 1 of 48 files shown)


12wrigja:
Thanks for this PR! I'll send it for a run through Google's internal CI to make sure it can land and it will be auto-merged once it lands internally.

12wrigja:
Internal CI points out there are some test failures that need addressing. Unfortunately, we don't have a straightforward test setup for the OSS repo, but the debugging steps shouldn't be too painful:

 - `npm install` if you haven't already
 - `npm run serve_tests`
 - open Chrome to the index page (`http://localhost:8080/` should work)
 - append `closure/goog/string/string_test.html` as the path to a test.
 
 From what I can tell, the following testcases fail in Chrome (and other browsers we have in our internal CI):

  - string/string_test.html testCaseInsensitiveEndsWith
  - dom/browserrange/browserrange_test.html testDiv

CommanderRoot:
> * string/string_test.html testCaseInsensitiveEndsWith
This was fixed by adding an empty suffix check. I also added one on the prefix function. It's not needed there but it should be faster and this way the functions still look nearly the same.

> * dom/browserrange/browserrange_test.html testDiv
I found the test but the functions called there are not part of this codebase so not sure what's failing there. Maybe they used the above function and failed because of that?

closure/goog/dom/browserrange/abstractrange.js
```diff
@@ -257,7 +257,7 @@ goog.dom.browserrange.AbstractRange.prototype.getHtmlFragment = function() {
           output.append(html);
         } else {
           var index = html.lastIndexOf('<');
-          output.append(index ? html.substr(0, index) : html);
+          output.append(index !== -1 ? html.slice(0, index) : html);
```
12wrigja, line 260: This change isn't right, and is what is causing the browserrange tests to fail. The original logic is intentional:
 - if index is -1, then this appends empty string.
 - if index is 0, then the entire HTML content should be added
 - if the index is > 0, then only the portion of the html after the last open tag is appended.
 
 Can you add a comment to help document this?

12wrigja:
I quickly stepped through this in a debugger and found the change that causes problems here, and left you a comment.

closure/goog/dom/browserrange/abstractrange.js
CommanderRoot: Good catch. I haven't seen it being done like this yet. I remember being confused about the `index ?` part when I changed it but I didn't think deeper about it.
I have added a comment about the wanted behaviour and also fixed the code to reflect that.

12wrigja: Thanks! This LGTM.

12wrigja:
I've synced this into Google and am re-running the CI. This should go through internal review in the next day or so - thanks for your patience.

12wrigja:
Merged this in with some small changes based on feedback in internal review. Thanks for your contribution!

Issue #1124, opened 2021-03-10
Wrong placement of negative sign in currency for Norwegian locale


NorseJedi:
I originally reported this as an Angular bug pertaining to the currency pipe, and was told to report it here.

The locale for Norway places the negative sign in front of the symbol rather than in front of the digits.

Negative currency values will currently be formatted like this:
```−kr 1 000,00```

The correct way to format them for Norwegian locale however, is this:
```kr −1 000,00```

12wrigja:
As the file overview comment indicates, this might be the result of data missing from the CLDR dataset: https://github.com/google/closure-library/blob/39cb429b2e3c12170270b1e92ae2fbaeec007d75/closure/goog/i18n/numberformatsymbols.js#L9

I would guess the correct file to get updated would be https://github.com/unicode-org/cldr/blob/367b79b4a86a9cb12314158897d950268dc77b02/common/main/nb.xml#L11516 but I'd recommend filing a bug against the CLDR data first: http://cldr.unicode.org/index/bug-reports

12wrigja:
I filed https://unicode-org.atlassian.net/browse/CLDR-14591 to see if this can get fixed upstream.

12wrigja:
A quick update in case you weren't subscribed to that issue: it seems that this is fixed in CLDR 40. I'm not sure when that will land in Closure Library (that process is managed by a different team at Google).

Akisan98:
Hi If I'm allowed I want to add that Currency should fall behind the numbers: 20 NOK or 20 Kr, instead of in front, like how its handled for French.

ensnared:
> Hi If I'm allowed I want to add that Currency should fall behind the numbers: 20 NOK or 20 Kr, instead of in front, like how its handled for French. "¤ #,##0.00"
> 
> #1157

This is incorrect.

While both placements are allowed both for "NOK" and "kr", the most common placement for "NOK" is in front of the number.
For "kr" it is usually placed after the number when it's part of a written sentence, while in lists and tables and such it can be either (although since MS Excel has always put it in front that's probably the most common placement).

Ref:
Norwegian original: https://www.sprakradet.no/svardatabase/sporsmal-og-svar/kronebelop-rekkjefolgje-komma-og-strek/
[...]

PR #1176, opened 2022-09-29
Fix zh locale i18n strings for 0 minutes and 0 hours

zac1st1k:
`这一时间 / 此时` contains two strings those are not really usable.

`此刻` means `now` is the same as 0 second translation `现在`

In mandarin, we don't say anything similar to "this minute" or "this hour". The added `刚刚` is commonly used meaning "moment ago" or "just now".

I cannot find a good translation for `this hour`, I used simplified `本小时` from `zh_Hant_HK`.

### DIFF
--- a/closure/goog/i18n/relativedatetimesymbols.js
+++ b/closure/goog/i18n/relativedatetimesymbols.js
@@ -9893,14 +9893,14 @@ exports.RelativeDateTimeSymbols_zh =  {
   },
   HOUR: {
     LONG:{
-      R:{'0':'这一时间 / 此时'},
+      R:{'0':'本小时'},
       P:'other{#小时前}',
       F:'other{#小时后}',
     },
   },
   MINUTE: {
     LONG:{
-      R:{'0':'此刻'},
+      R:{'0':'刚刚'},
       P:'other{#分钟前}',
       F:'other{#分钟后}',
     },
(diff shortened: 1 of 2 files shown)


12wrigja:
Unfortunately, we cannot accept this PR because the file you are changing is generated from a CLDR dataset - see my explanation in https://github.com/google/closure-library/issues/1124#issuecomment-796346176.

If you'd like to correct these symbols, the best thing to do is to open a ticket against the dataset to ask them to survey the change and see if it can be merged in. Once a CLDR version is released that includes the changes, we will automatically update these files accordingly.

zac1st1k:
Thanks @12wrigja 
According to https://unicode-org.github.io/cldr-staging/charts/latest/verify/dates/zh.html latest CLDR chart. It has `this year` and `this month` etc but it doesn't have `this hour` or `this minute`.
May I know where the dataset comes from?

12wrigja:
> May I know where the dataset comes from?

I'm not sure what you are asking for here - the dataset is curated manually by the Unicode organization, and we pull the dataset into Closure within Google and generate JS files from the data inside.

If the upstream dataset is entirely missing a field for "this hour" or "this minute" then that is something to raise with the CLDR maintainers.

I'm going to close this, because we cannot accept changes to these files given they are generated from upstream datasets.

Issue #1214, opened 2023-11-01
Closure Library is in Maintenance Mode


shicks:
***TL;DR: Closure Library will be sunset over the next 9 months, from November 1, 2023, to August 1, 2024. We no longer see it as meeting the needs of modern JavaScript development, and we recommend that users look for alternative solutions. Feel free to skip the background and go straight to the FAQ.***

---

In the nearly 14 years since Closure Library was first made public, it's broken a lot of ground and paved the way for numerous ecosystem features that we take for granted today.  At the initial release in 2009, Internet Explorer 8 was brand new, Chrome had just released version 3, the iPhone 3GS was fresh, and ECMAScript 5 was still just around the corner.  Browser implementations varied widely in basic language and DOM support, with very different event models and quirks around every corner.  Dependencies were generally managed manually by loading scripts in the right order.

Closure Library brought some amazing innovations to this ecosystem by providing a common library that glossed over browsers' many differences and allowed authors to focus on features and logic, rather than compatibility and quirks.  It provided a full-featured debug loader, a wide swath of versatile tools, and a large gallery of prepackaged composable widgets that anybody could use to build web apps at a scale previously unimagined.  And the type checker provided a level of static analysis that was previously unheard of in JavaScript.

But these same innovations that were groundbreaking at the time are now starting to feel a bit old and dated.  CommonJS and ES Modules are the gold standard for dependency management; the language has evolved and added new, safer, and more ergonomic APIs for many basic features like classes or events; we now have TypeScript that provides a far more expressive type system than Closure; and groups like WICG and TAG have substantially reined in the divergences between different browsers.  **In short, the world has moved on and we believe it no longer needs Closure Library.**  These days, rather than elevating a project to modern practices, a dependency on Closure Library seems to feel more like something weighing a project down, holding it back from modern standards and greater adoption.  In light of this, we've made the difficult decision that it's time to thank Closure Library and its many authors and contributors for all they've done for us over the years, and to bid it farewell.

# FAQ

## What is happening?

On November 1, 2023, we will transition Closure Library into maintenance mode.  After this date, we will only push critical commits to the GitHub repository, respond to critical issues, and merge critical pull requests.  On August 1, 2024, we plan to permanently freeze and archive the GitHub repository, after which no new commits or releases will be pushed at all.

During this time, we recommend looking for alternative solutions so that dependent projects can continue to receive critical security updates into the future.

## Why are we doing this?

[...]

PR #1226, opened 2024-04-26
Update article.js DOM text reinterpreted as HTML

Shivam7-1:
By using textContent, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.


trevorade:
Please see https://github.com/google/closure-library/issues/1214.

TL;DR: Closure Library is in maintenance mode, and we aren't accepting PRs at this time that aren't considered critical.

